PR #2, opened 2016-02-19
enddat-122; initial commit for new enddat project structure

mamcderm-usgs:


bower.json
```diff
@@ -0,0 +1,34 @@
...
+		"bootstrap": "https://github.com/twbs/bootstrap.git#3.3.5",
+		"text": "requirejs-text#~2.0.14",
+		"underscore": "~1.8.3",
+		"backbone": "~1.2.3",
+		"handlebars": "~4.0.3",
+		"sinon": "https://github.com/sinonjs/sinon.git#1.17.0",
+		"Squire.js": "https://github.com/iammerrick/Squire.js.git#0.2.0",
+		"font-awesome": "fontawesome#~4.4.0",
+		"requirejs-hbs": "~0.1.1",
+		"jquery-ui": "~1.11.4"
```
mbucknell: I don't think we need jquery-ui. If not, this should be removed.

pom.xml
```diff
@@ -219,66 +141,35 @@
...
 		<dependency>
-			<groupId>org.webjars</groupId>
-			<artifactId>jsts</artifactId>
-			<version>${jsts.version}</version>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-simple</artifactId>
+			<version>1.7.12</version>
 		</dependency>
-		<!-- </webjars> -->
 	</dependencies>
```
mbucknell: There are still dependencies that I don't think we need. Such as jdbc spec. Do we need tomcat filters (is this used by the proxy code)?

jiwalker-usgs: tomcat filters is used to read some custom headers that are passed by the apache proxy in order to reconstruct the URL as requested by the user.  It the replaces the `request.getRequestUrl()` and other calls so the code doesn't need to know about the headers.

jiwalker-usgs: That said, it is probably not needed.

mbucknell: @jiwalker-usgs, thanks for the explanation. We are trying to start with exactly what we need at the moment so, yes, I think it should be removed until we think we need it.

src/main/webapp/dataDiscovery.jsp
```diff
@@ -0,0 +1,78 @@
...
+			<jsp:param name="description" value="" />
+			<jsp:param name="author" value="" />
+			<jsp:param name="keywords" value="" />
+			<jsp:param name="publisher" value="" />
+			<jsp:param name="revisedDate" value="" />
+			<jsp:param name="nextReview" value="" />
+			<jsp:param name="expires" value="never" />
+			<jsp:param name="development" value="false" />
+		</jsp:include>
+		<link rel="stylesheet" type="text/css" href="css/nude.css">
```
mbucknell: I think we should remove this style sheet and the one on the line below, unless their removal affects the header and footer. Going forward we will need to take a closer look at the style sheets used for the different pages in this project.

mbucknell:
We should remove the CIDA_Logo.png from the repo and remove the image from any page where it is displayed.

mbucknell:
We aren't using less yet for our style sheets. We will need to do that soon, but let's worry about that later.

I think you can go ahead and merge. You can address the comments later.

PR #3, opened 2016-02-22
frontend-maven-plugin now runs karma rather than npm test for test st…

mbucknell:
…ep. Updated karma.conf.js to include all flags needed in the CI and build environment. To make things easier on a developer modified the test script to run without a browser, with single-run disabled, and reporters progress. This assumes that npm is installed on the caller's environment. This is merely a convience for the developer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "EnDDaT User Interface",
   "main": "dataDiscovery.jsp",
   "scripts": {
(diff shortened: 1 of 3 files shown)


mbucknell:
@mamcderm-usgs , I'm going to go ahead and merge this. I'd like you to try building on your windows machine and using the npm test command from the command line to run the tests with single-run disabled.

PR #12, opened 2016-02-26
Enddat 134

mbucknell:
The NavView will need to have more functionality eventually to do things like set the state of the buttons, click handlers, etc. Right now there is nothing to test in there so I didn't add any tests.

I also am defining less files to go along with the handlebars templates. Not sure if I like this yet, but that is the direction I'm taking at the moment.


mamcderm-usgs:
:+1:

PR #13, opened 2016-02-29
Enddat 135

mbucknell:
Brought in Leaflet and Leaflet.Providers to implement the map and base layers. Added a MapView which will create a map with base layers and a layer switcher control. Added tests

Brought in the karma-sinon framework and removed the sinon dependency from bower.json.

Reformatted bower.json and karma.conf.js to use tabs.

### DIFF
--- a/src/test/js/karma.conf.js
+++ b/src/test/js/karma.conf.js
@@ -1,95 +1,69 @@
 // Karma configuration
 // Generated on Tue Feb 09 2016 11:24:56 GMT-0700 (Mountain Standard Time)
 
-module.exports = function(config) {
-  config.set({
-
-    // base path that will be used to resolve all patterns (eg. files, exclude)
-    basePath: '../../..',
-
-
-    // frameworks to use
-    // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
-    frameworks: ['jasmine', 'requirejs', 'sinon'],
-
-
-    // list of files / patterns to load in the browser
-    files: [
-        'src/test/js/test-main.js',
-      {pattern: 'src/main/webapp/bower_components/Squire.js/src/Squire.js', included: false},
-      {pattern: 'src/main/webapp/bower_components/jquery/dist/jquery.js', included: false},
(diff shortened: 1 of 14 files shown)


src/test/js/karma.conf.js
```diff
@@ -1,94 +1,69 @@
...
-    concurrency: Infinity
-  })
-}
+module.exports = function (config) {
+	config.set({
+		// base path that will be used to resolve all patterns (eg. files, exclude)
+		basePath: '../../..',
+		// frameworks to use
+		// available frameworks: https://npmjs.org/browse/keyword/karma-adapter
+		frameworks: ['jasmine', 'requirejs', 'sinon'],
```
mbucknell: Added the sinon framework. So we don't need to pull in that dependancy with bower or specify it in the require setup or in the files to include.

src/test/js/karma.conf.js
```diff
@@ -1,94 +1,69 @@
...
+			'src/test/js/test-main.js',
+			{pattern: 'src/main/webapp/bower_components/Squire.js/src/Squire.js', included: false},
+			{pattern: 'src/main/webapp/bower_components/jquery/dist/jquery.js', included: false},
+			{pattern: 'src/main/webapp/bower_components/underscore/underscore.js', included: false},
+			{pattern: 'src/main/webapp/bower_components/backbone/backbone.js', included: false},
+			{pattern: 'src/main/webapp/bower_components/bootstrap/dist/js/bootstrap.js', included: false},
+			{pattern: 'src/main/webapp/bower_components/handlebars/handlebars.amd.js', included: false},
+			{pattern: 'src/main/webapp/bower_components/text/text.js', included: false},
+			{pattern: 'src/main/webapp/bower_components/requirejs-hbs/hbs.js', included: false},
+			{pattern: 'src/main/webapp/bower_components/leaflet/dist/leaflet.js', included: false},
```
mbucknell: Added leaflet.js and leaflet-providers.js

mamcderm-usgs:
looks great, thanks for the comments :+1:

PR #14, opened 2016-03-01
Enddat 136

mbucknell:
Added the ability to click on a location on the map, create a marker, and save the location to a model.

As part of this ticket, I also updated the NavView extensively to start implementing the navigation. I added routes for the choose data step (although I have only tested the first). Note that the url is updated as the workflow state changes but this does not render the entire page.

I also added loglevel to provide javascript logging.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,35 +1,35 @@
 {
-  "name": "enddat_web",
-  "version": "0.0.0",
-  "homepage": "https://github.com/enddat_web",
-  "moduleType": [
-    "amd"
-  ],
-  "license": "MIT",
-  "ignore": [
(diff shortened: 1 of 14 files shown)


mbucknell:
@mamcderm-usgs , this pull request is not yet ready for review, but I thought you might want to see how I've implemented the WorkflowStateModel and how I'm using that information. If it's helpful to have the changes I have made to add the WorkflowStateModel, then we can go ahead and merge. The tests are passing.

mbucknell:
Going to go ahead and merge this. When I make the final pull request for this ticket, I will try to point out where changes were made.

PR #15, opened 2016-03-02
Enddat 136

mbucknell:
This was mostly adding tests for NavView and the new functionality in MapView. The routing was also fixed and now works correctly with the optional parameters for the choose data step. This pull request should be reviewed with #14 .


mamcderm-usgs:
looks good to me :+1:

PR #16, opened 2016-03-03
Enddat 138

mamcderm-usgs:


mbucknell:
Mark could you use git hub comments to point out the places where you want me to look?

mamcderm-usgs:
@mbucknell - little bit of a catch-22 right now, I'm making progress on some of the issues, so the code is in flux.  I'm getting nervous about my progress because I don't want to hold you up but maybe best to wait until I have another push.

src/main/webapp/WEB-INF/jsp/head.jsp
```diff
@@ -19,10 +19,11 @@
 	String baseUrl = props.getProperty("enddat.base.url", request.getContextPath());
 	if (!baseUrl.endsWith("/")) { baseUrl += "/"; }
 	String version = props.getProperty("application.version");
+	String proxyServiceUrl = props.getProperty("enddat.proxy-service-url");
```
mamcderm-usgs: Not sure if this is needed since it's the same as baseUrl at this point.

src/main/webapp/dataDiscovery.jsp
```diff
@@ -27,8 +27,13 @@
 				config: {
 					'init': {
 						'contextPath': "<%=baseUrl%>",
-						'development' : '<%=development%>'
-					}
+						'development': '<%=development%>'
+					},
+					//this doesn't seem to work, not sure how else to get jndi variables in
+					'SiteModel' : {'proxyUrl': "<%=proxyServiceUrl%>"},
+					'ParameterCodes' : {'proxyUrl': "<%=proxyServiceUrl%>"},
```
mamcderm-usgs: I would think this would work to bring in a context variable through module but it doesn't seem to work.  Is this the right idea or is there a better approach?

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,89 @@
...
+	"use strict";
+
+	//does this need to be a collection?
+	var model = Backbone.Model.extend({
+		url: 'waterService/?format=rdb&bBox=' +
+		//still need to bring over code to get these values from passed in lat, lon, rad
+		-105.213267 + ',' +
+		39.646356 + ',' +
+		-105.025118 + ',' +
+		39.791079 +
```
mamcderm-usgs: Is there any reason to not use the old code to create the parameters for the bounding box?

src/main/webapp/WEB-INF/jsp/head.jsp
mbucknell: I would say it's not needed.

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,89 @@
...
+				"parm_cd" : null,
+				"stat_cd" : null,
+				"loc_web_ds" : null,
+				"begin_date" : null,
+				"end_date" : null,
+				"count_nu" : null
+			};
+			
+			var lines = data.split("\n");
+			parseRDB(lines, importantColumns, function(colVals) {
```
mamcderm-usgs: Parse does not seem to be getting fired on the fetch, is there something wrong here?

src/main/webapp/dataDiscovery.jsp
mbucknell: Should work ... I don't see where you are calling module.config() in either of those modules?

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -21,6 +22,16 @@ define([
 		 *		@prop {models/WorkflowStateModel} model
 		 */
 		initialize: function (options) {
+			
+			if (this.model.attributes.step === this.model.CHOOSE_DATA_STEP) {
+				//load the site model based on the properties in this.model
+				this.siteData = new SiteModel({},{projectModel: this.model});
+
+				//probably need to call function with this.mapView in done 
+				this.siteData.fetch().done();
```
mamcderm-usgs: Parse() does not seem to get fired on this fetch and I haven't figured out why.  Also, assuming that does work properly, is the right idea to put the mapView creation in a separate function in the view and call it from done()?

src/main/webapp/js/models/SiteModel.js
mbucknell: If it works, its fine.

src/main/webapp/js/models/SiteModel.js
mbucknell: I don't see anything obvious. You will need to step into the fetch and see what is happening.

src/main/webapp/js/views/DataDiscoveryView.js
mbucknell: I think the better thing to do is to pass in siteData to mapView and then have mapView listen to events on the siteData model. Specifically, there is a 'sync' event that is fired after successful fetching of the data. http://backbonejs.org/#Events-catalog.

src/main/webapp/dataDiscovery.jsp
mamcderm-usgs: I took out the module.config().proxyUrl in those modules because it wouldn't resolve and I couldn't figure out why.

mamcderm-usgs: It works without the proxyUrl but I was still trying to understand how to get it to work for future reference.

mbucknell: At second glance, it's because you need to use the path from baseUrl, so it would be models/SiteModel , etc

mamcderm-usgs: of course, so obvious...thank you.

src/main/webapp/js/models/ParameterCodes.js
```diff
@@ -0,0 +1,40 @@
...
+], function ($, Backbone, ParseRDB, module) {
+	"use strict";
+	
+	//should this be a collection rather than a model?
+	var model = Backbone.Model.extend({
+		url: 'pmcodes?radio_pm_search=param_group&pm_group=Physical&format=rdb&show=parameter_nm',
+		
+		//still not sure if there should be an init with fetch() or do fetch in router?
+		initialize: function() {
+			this.fetch();
```
mbucknell: My opinion is that the fetch should be in the router. I don't like adding this kind of active code to the initialize. It makes testing more difficult.

src/main/webapp/js/views/DataDiscoveryView.js
mbucknell: parse will only get called if the fetch is successful. I would check the status of the promise or add a fail temporarily with a log message so that you can see if it is succeeding.

mamcderm-usgs: Yes, I put an error handler on the fetch and it is not successful (even though the ajax call is).  It's not successful because the response is not json, I think.  So, I think I need to setup a collection and use it's parse to create the model, with proper json.  I have yet to really fully comprehend the difference between model and collection.  If I have a table like this:   
row | column 1 | column 2
1     |      a         |  a description
2     |      b         |  b description

I think I can have a model like {a : a decription, b : b description}  or a model like this {column 1 : column 2}.
Using the second model, I can  have a collection like this: {model 1, model 2}
Either way, I think I can access the key value pairs, as long as the keys on the first model are different.  So, what is the appropriate criteria for when to use models vs collections?  Answer - use models if unique keys and you don't need to parse the url response to get proper json, otherwise use collections?

mbucknell: I think you should use collections when the data is an array of objects (which are models). And it does seem more appropriate in this case to make this a collection of models. Each object in the model will at least contain an id, location, and whatever other information comes along. So I think each row as you described above would be a model in the collection.

mamcderm-usgs: I guess a more general thought is to use a backbone collection when modeling a table and just use a backbone model when modeling something not necessarily on a table, like actions on the UI.  I was definitely getting tripped up on this so hopefully I can make better progress now.  Thanks.

mbucknell: And just to be clear, each row in a table is a model. In  My pubs, the publications where a collection of publication models. The pages to edit a publication, fetch a publication model.

mamcderm-usgs:
@mbucknell - This is still rough but wanted to get latest up in case you could review and guide.  The code doesn't throw any errors, but I still need to make sure I have the correct model structure, load the model/collection correctly, setup the mapView to listen to changes on the SiteModel, setup proper error handling, create tests.  Every task I am still learning/reinforcing fundamentals so it is painfully tedious.  Let me know if you would be willing to pair code/tutor to help me accelerate learning curve.  Thanks.

src/main/webapp/WEB-INF/jsp/head.jsp
```diff
@@ -22,7 +22,7 @@
 %>
 
 <link rel="shortcut icon" type="image/ico" href="img/favicon.ico">
-<jsp:include page="template/USGSHead.jsp">
+<jsp:include page="/template/USGSHead.jsp">
```
mbucknell: Not sure why you changed this. Was it not working?

src/main/webapp/js/models/ParameterCodes.js
```diff
@@ -0,0 +1,51 @@
...
+			var self = this;
+			$.ajax({
+				type : "GET",
+				url : self.url,
+				dataType: 'text',
+				success: function(data) {
+					self.parse(data);
+				},
+				error : function(jqXHR, textStatus, error) {
+					$('#errorMessage').html("Error in loading NWIS Parameter definitions: " + textStatus);
```
mbucknell: There should be no DOM manipulation code in the models. You can pass in an error handler function into the fetch call or handle the promise that is returned from the fetch. The fetch should return the promise that is produced by the $.ajax call.

src/main/webapp/js/models/ParameterCodes.js
```diff
@@ -0,0 +1,51 @@
...
+					self.parse(data);
+				},
+				error : function(jqXHR, textStatus, error) {
+					$('#errorMessage').html("Error in loading NWIS Parameter definitions: " + textStatus); 
+				}
+			});
+		},
+
+		parse: function(data) {
+			var NWIS_PARAMETER_CODE_DEFINITIONS = {};
```
mbucknell: Only hard coded constants should use all upper case. This is being filled in by the call so should use camelCase

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,122 @@
+/* jslint browser: true */
```
mbucknell: Consider renaming this file to SiteCollections (or something with Collections in the name).

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,122 @@
...
+	'util/utils',
+	'util/geoSpatialUtils',
+	'models/ParameterCodes',
+	'models/StatisticCodes'
+], function ($, Backbone, utils, geoSpatialUtils, ParameterCodes, StatisticCodes) {
+	"use strict";
+
+	DEFAULT_RADIUS : 5;
+
+	var model = Backbone.Model.extend({});
```
mbucknell: This seems unnecessary.

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,122 @@
...
+				"begin_date" : null,
+				"end_date" : null,
+				"count_nu" : null
+			};
+			
+			var lines = data.split("\n");
+			utils.parseRDB(lines, importantColumns, function(colVals) {
+				var site;
+				//Add the info to the sites
+				if (!sites.hasOwnProperty(colVals["site_no"])) {
```
mbucknell: Consider using underscore's '_.has', rather than .hasOwnProperty.

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,122 @@
...
+					site = {};
```
mbucknell: Could initialize with a single assignment statement.

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,122 @@
...
+					site.name = colVals["station_nm"];
+					site.lat = colVals["dec_lat_va"];
+					site.lon = colVals["dec_long_va"];
+					site.parameters = [];
+				} else {
+					site = sites[colVals["site_no"]];
+				}
+
+				var name = "Unknown parameter " + colVals["parm_cd"] + " " + colVals["stat_cd"];
+				if (ParameterCodes && StatisticCodes) {
```
mbucknell: Is this part of the code working? ParameterCodes and StatisticCodes are model objects not instances of the models. I think you will need to pass in the instances of each of these models.

src/main/webapp/js/models/StatisticCodes.js
```diff
@@ -0,0 +1,53 @@
...
+			var self = this;
+			$.ajax({
+				type : "GET",
+				url : self.url,
+				dataType: 'text',
+				success: function(data) {
+					self.parse(data);
+				},
+				error : function(jqXHR, textStatus, error) {
+					$('#errorMessage').html("Error in loading NWIS stat definitions: " + textStatus);
```
mbucknell: Same comment as with ParameterCodes.

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,122 @@
...
+				success: function(data){
+					self.parse(data);
+				},
+				error: function(jqXHR, textStatus, errorThrown) {
+					//Do we have to clear the NWIS tab here? or is it done somewhere else?
+					if (404 === jqXHR.status) {
+						//No data, do nothing
+					} else {
+						//Couldn't get to NWIS Web
+						$('#errorMessage').html("Error in loading NWIS data: " + textStatus);
```
mbucknell: Same comment as ParameterCodes

src/main/webapp/js/models/StatisticCodes.js
```diff
@@ -0,0 +1,53 @@
...
+				},
+				error : function(jqXHR, textStatus, error) {
+					$('#errorMessage').html("Error in loading NWIS stat definitions: " + textStatus); 
+				}
+			});
+			
+		},
+
+		parse: function(data) {
+			var NWIS_STAT_CODE_DEFINITIONS = {};
```
mbucknell: Use camelCase.

src/main/webapp/js/util/utils.js
```diff
@@ -0,0 +1,52 @@
...
+define([
+	'underscore',
+	'jquery'
+], function (_, $) {
+	"use strict";
+	
+	var self = {};
+
+	self.parseRDB = function(lines, importantColumns, onRowCallback) {
+		var columnIndexes = $.extend({}, importantColumns);
```
mbucknell: Rather the bringing in jquery, let's use the functions in underscore for extend and each.

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -23,6 +24,17 @@ define([
...
 		initialize: function (options) {
+			
+			if (this.model.attributes.step === this.model.CHOOSE_DATA_STEP) {
+				//load the site model based on the properties in this.model
+				this.siteData = new SiteModel({},{projectModel: this.model});
+
+				//pass in siteData to mapView and then have mapView listen to events
+				//on the siteData model. Specifically, there is a 'sync' event that is
+				//fired after successful fetching of the data.
+				this.siteData.fetch();
```
mbucknell: The fetch should only happen if the location, radius, and dataset set to 'NWIS' are in the workflow model. Then we should set up event handlers on the workflow model to monitor changes to these attribute and update this.siteData accordingly. This means if dataset does not include NWIS, the siteData model should be cleared.

src/main/webapp/WEB-INF/jsp/head.jsp
mamcderm-usgs: It doesn't work in my environment unless I put in a "/", I have no idea why.  After a long time, I gave up trying to figure it out but I'm sure there's a trivial explanation.

src/main/webapp/js/models/ParameterCodes.js
mamcderm-usgs: I understand.  I just copied in this code from the old enddat and I'm trying to understand how to get the basic model/collection mechanism to work with a non-JSON response, which is unfortunately eluding me.  Once I get the basic model to work, I was planning to go back and translate/refine how this should work within this project.

mbucknell: I did add a logging framework which would be safer to use while you are writing code in case this gets left in accidentallly. If you use log.debug, those messages will only show up in the console log if the development variable in the context is set to true

src/main/webapp/js/models/SiteModel.js
mamcderm-usgs: Not sure what you're referring to?

mbucknell: You don't need to include the model property when you extend the collection if you are using the Backbone.Model without extending it at all. That's what the model property is set to be default.

src/main/webapp/js/models/SiteModel.js
mamcderm-usgs: Yeah, I understand that too.  Originally, I was instantiating in the model then passed the instance this way, not sure how I need to do this yet with the fetch moved to the router.  At this point still focusing on trying to get the collection/model to load from url still, so will have to revisit this.

src/main/webapp/js/models/ParameterCodes.js
```diff
@@ -0,0 +1,33 @@
...
+				},
+				error : function(jqXHR, textStatus, error) {
+					$('#errorMessage').html("Error in loading NWIS Parameter definitions: " + textStatus); 
+				}
+			});
+		},
+
+	});
+
+	return collection;
```
mamcderm-usgs: @mbucknell - I've contrived this function to see if I can load a collection based on data coming from a fetch override and this still doesn't seem to load the models into the collection.  The best stackoverflow I can find on the issue is this:  http://stackoverflow.com/questions/23722137/backbone-working-with-non-json-data

Can you see what is wrong or what I can try to get the collection loaded?

mbucknell: On line 21 you are creating an array of plain objects ... not model instances. So you have to do something like this (using your example):
arrayOfModels =[new Backbone.Model({'00001':'parm1'}), new Backbone.Model({'00002':'parm2'})]

For reference: http://backbonejs.org/#Collection-set

mamcderm-usgs: That doesn't seem to work either.  I was just trying set() out of desperation and it was mentioned in that stackoverflow.  There seems to be some issue with overriding fetch/parse that gets in the way of creating a model/collection.  I know we want to use a backbone model for this but I am running out of ideas for how to get this to work.

mbucknell: See https://github.com/USGS-CIDA/enddat_web/pull/19

mbucknell:
@mamcderm-usgs , I think you should probably close this pull request. It really was just for reviewing code.

mamcderm-usgs:
@mbucknell - I pushed the latest code.  I still need to setup the listener and write tests so this is available in case you have time to review before I get those done.  I setup the parameter codes and statistic codes as backbone collections, as we discussed, but I'm still not sure that a backbone model wouldn't be simpler/better for those.  Please send any other suggestions you have for the promises, error handling, fetches, etc.

mbucknell:
@mamcderm-usgs, could you close this pull request and open a new one? That way we can start fresh.

PR #17, opened 2016-03-03
Enddat 137

mbucknell:
Added a box to allow the user to specify or change the location. This is in a collapsible panel. I created a BaseCollapsiblePanelView as there will be more of these. To create a specific panel, you will extend the BaseCollapsiblePanelView. I am using this for the LocationView.

Added geoSpatialUtils and put the code from the old enddat application in here that calculates the bounding box from a location and radius. I'm sure this package will grow.


bower.json
```diff
@@ -1,35 +1,36 @@
 {
```
mbucknell: When you add files to bower with the --save-dev, it takes out the tabs and uses spaces, so I think we can just leave this one to use spaces.

src/main/webapp/js/views/NavView.js
```diff
@@ -1,11 +1,12 @@
 /* jslint browser : true */
 
 define([
+	'underscore',
 	'bootstrap', // Needed by the bootstrap navbar
 	'loglevel',
 	'views/BaseView',
-	'hbs!hb_templates/workflowNav'
-], function(bootstrap, log, BaseView, hb_template) {
+	'hbs!hb_templates/workflowNav',
```
mamcderm-usgs: extra comma

mamcderm-usgs:
Wow!  Looks great.  :+1:

PR #19, opened 2016-03-04
Added ParameterCodes and code to AppRouter to load them.

mbucknell:
@mamcderm-usgs , I added your ParameterCodes package, changed to create model instances and returning the promise. I am loading the parameter codes in the app router's initialize and it is loading into the collection. I hope this helps.


mamcderm-usgs:
So, I was looking at parameterCodes in the chrome tools debugger and it doesn't have the models.  I guess that must be some async issue because the log message indicates there are 3.  So, I guess I wasted a couple days or so on this....

mbucknell:
If you put a breakpoint in the app router in the fetch.done (line 26), you should be able examine self.parameterCodes and see that there are indeed three Backbone.Models.

mamcderm-usgs:
hmmm, that's what I've been doing but I still don't see them.  I'm not sure if I'm just not looking at it right or I've got some other problem, but the console message does say there are 3.

PR #20, opened 2016-03-08
Enddat 138

mamcderm-usgs:


src/main/webapp/js/controller/AppRouter.js
```diff
@@ -17,8 +19,18 @@ define([
...
 			this.workflowState = new WorkflowStateModel();
+
+			this.parameterCodes = new ParameterCodes();
+			this.pCodesPromise = this.parameterCodes.fetch().done(function() {
+				log.debug('Fetched parameter codes ' + self.parameterCodes.length);
+			});
+			this.statisticCodes = new StatisticCodes();
+			this.sCodesPromise = this.statisticCodes.fetch().done(function() {
+				log.debug('Fetched statistic codes ' + self.statisticCodes.length);
+			});
```
mamcderm-usgs: I'm not sure if this is the best place to instantiate these models.  If the app only uses these in conjunction with the site model, then I'm guessing we should do it in there.  Leaving them here for now though until I learn more.

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -1,17 +1,20 @@
...
 define([
+    'loglevel',
 	'views/BaseView',
 	'views/NavView',
 	'views/MapView',
 	'views/LocationView',
+	'models/SiteCollection',
 	'hbs!hb_templates/dataDiscovery'
-], function (BaseView, NavView, MapView, LocationView, hbTemplate) {
+], function (log, BaseView, NavView, MapView, LocationView, SiteCollection, hbTemplate) {
```
mamcderm-usgs: should log go in baseView?

src/main/webapp/js/controller/AppRouter.js
mbucknell: What do you think about not making these separate models? I think you can make all three calls in SiteModel. Of course you would only make the ajax calls to get the parameter and statistics codes once (either in the initialize or in the first fetch).

mamcderm-usgs: Are you suggesting combining parameter codes and statistic codes into one model?  Would the attributes be something like this:

{parameter_codes: [parameter code objects, ex. {parm_cd: 00001, parm_nm: "water temp"}, {etc.}], statistic_codes: [statistic code objects]}

If I have this right it seems simpler to keep them separate.  I do agree we should probably instantiate/fetch these models in SiteModel, of course unless they are used in other views but I'm not thinking that is the case.  Another consideration may be whether or not to change the SiteModel to not resolve the parameter code names and statistic code names in the SiteModel.  Alternatively, the names could be looked up on a Singleton model in the view.  Not sure if that makes sense though.

src/main/webapp/js/views/DataDiscoveryView.js
mbucknell: I don't think so. That would just obscure where the log function comes from. This makes it explicit where log is used. Also if you put it into BaseView every view would require it and that seems unnecessary.

src/main/webapp/js/controller/AppRouter.js
mbucknell: Actually, what I'm suggesting is that they shouldn't be represented by Backbone.Model at all. Rather these would be attributes within the SiteModel as plain json objects. So your SiteModel would have a parameterCodes attribute and a statisticCodes attribute.

As to your last paragraph, I'll look at that more closely as I review the code more closely.

mamcderm-usgs: Okay, that does seem cleaner.  I had the same thought at one point.  It is just a very simple lookup so an attribute with a json object as it's value on the Site Model would seem to be just right.  I'll wait to hear from you but let me know if you want me to go with that change.  I think I should be able to do that pretty quick.  The other possible change I mentioned regarding the names can be considered separately.

src/main/webapp/js/views/BaseView.js
```diff
@@ -50,6 +50,11 @@ define([
 			}
 
 			Backbone.View.prototype.initialize.apply(this, arguments);
+		},
+
+		/** COMMON navigation functions that might be used for many views */
+		goHome: function() {
+			this.router.navigate('', {trigger: true});
```
mbucknell: You can make this more flexible by passing in the options to the navigate function. In many cases, we are not going to be setting trigger to true, for example.

src/main/webapp/js/controller/AppRouter.js
```diff
@@ -67,11 +79,14 @@ define([
 				'datasets' : datasets ? datasets.split('/') : null
 			});
 			this.createView(DataDiscoveryView, {
-				model : this.workflowState
-			}).render();
+				model : this.workflowState,
+				pCodesPromise : this.pCodesPromise,
+				sCodesPromise : this.sCodesPromise,
+				parameterCodes : this.parameterCodes,
+				statisticCodes : this.statisticCodes
```
mbucknell: Another bonus to fetching the parameter and staticistics codes in the SiteCollection is that we don't need to pass them into the DataDiscoveryView.

mamcderm-usgs: Definitely!

src/main/webapp/js/views/BaseView.js
mamcderm-usgs: So, is there any benefit to setting up a function like this or should we just call this.router.navigate() directly?

src/main/webapp/js/models/SiteCollection.js
```diff
@@ -0,0 +1,136 @@
...
+							site.parameters = [];
+						//another line with same site_no
+						} else {
+							site = siteElement[colVals["site_no"]];
+						}
+
+						var name = "Unknown parameter " + colVals["parm_cd"] + " " + colVals["stat_cd"];
+						if (options.parameterCodes && options.statisticCodes) {
+							if (colVals["parm_cd"]) {
+								var pCode = options.parameterCodes.find(function(model) {
```
mbucknell: Also making parameterCodes and statisticsCodes as json objects, it should eliminate the use of find. Should be able to just use _.has to see if the property is in the object.

src/main/webapp/js/models/SiteCollection.js
```diff
@@ -0,0 +1,136 @@
...
+						 * for each unique site_no, aggregating parameter codes.
+						 */
+						//first line
+						if(null == lastSite) {
+							siteElement = {};
+							lastSite = colVals["site_no"];
+						}
+						//after first line when new site_no, push last site model onto collection
+						else if(colVals["site_no"] != lastSite) {
+							siteCollection.push(new Backbone.Model(siteElement));
```
mbucknell: It's confusing to be pushing the last siteElement built onto the siteCollection. Can this be done at the end of the function? Also this seems to miss the last line.

src/main/webapp/js/models/SiteCollection.js
```diff
@@ -0,0 +1,136 @@
...
+						}
+						//after first line when new site_no, push last site model onto collection
+						else if(colVals["site_no"] != lastSite) {
+							siteCollection.push(new Backbone.Model(siteElement));							
+							siteElement = {};	
+							lastSite = colVals["site_no"];
+						}
+						//first time with new site_no
+						if (!_.has(siteElement, (colVals["site_no"]))) {
+							site = {};
```
mbucknell: Consider using _.groupBy to collect lines with the same site number and then process them together.

src/main/webapp/js/models/SiteCollection.js
```diff
@@ -0,0 +1,136 @@
...
+							startDate : colVals["begin_date"],
+							endDate : colVals["end_date"],
+							count : colVals["count_nu"]
+						});
+
+						siteElement[colVals["site_no"]] = site;
+						
+					});
+					//add last unique site_no model to collection
+					siteCollection.push(new Backbone.Model(siteElement));
```
mbucknell: See the comment above about doing this as the last step in the function

src/main/webapp/js/models/SiteCollection.js
```diff
@@ -0,0 +1,136 @@
...
+	'util/geoSpatialUtils'
+], function ($, Backbone, log, utils, geoSpatialUtils) {
+	"use strict";
+
+	var collection = Backbone.Collection.extend({
+
+		url: '',
+		
+		initialize: function(attributes, options) {
+			var workflowModel = options.workflowModel;
```
mbucknell: I think we should not be passing the workflowModel in. Instead, the url should be formed when fetch is called. Fetch can be called more than once. For instance if the workflow model state location is changed, fetch would be called with the new bounding box. I'll comment more about this in dataDiscoveryView.

src/main/webapp/js/models/SiteCollection.js
mamcderm-usgs: Yes.  Although this particular situation is what I was wondering if we really needed to do here.  I don't know enough at this point about when/where we need the parameter_nm and stat_nm, but it seems like it might be just as easy to pick them off the json object later, rather than stuff them in the Site Model?  Anyway, not critical to decide now but just a thought.

src/main/webapp/js/util/utils.js
```diff
@@ -0,0 +1,51 @@
+/*jslint browser: true */
+
+define([
+	'underscore'
+], function (_) {
+	"use strict";
+	
+	var self = {};
+
+	self.parseRDB = function(lines, importantColumns, onRowCallback) {
```
mbucknell: This function needs to have doc comments to describe inputs. I'm not a huge fan of callbacks. Consider just passing back an array of columns and letting the calling function process it.

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -31,11 +66,16 @@ define([
...
 
+			//pass in siteData to mapView where mapView listen to events
+			//on the siteData model. Specifically, there is a 'sync' event that is
+			//fired after successful fetching of the data.
 			this.mapView = new MapView({
 				el : this.$(MAPVIEW_SELECTOR),
 				mapDivId : 'map-div',
-				model : this.model
+				model : this.model,
+				sites : this.siteData
```
mbucknell: As part of this ticket, you should be adding the site markers to the map view.  Sorry if that wasn't clear.

mbucknell:
There are still a fair amount of work here. Specifically handling the model change events in DataDiscoveryView and drawing the sites in the MapView.

src/main/webapp/js/controller/AppRouter.js
mbucknell: I like resolving the parameter code and statistic code names within the SiteCollection.

src/main/webapp/js/models/SiteCollection.js
mbucknell: But the json object won't be exposed. It actually will because we will have to put it on "this", but I think I prefer completely parsing the response here. We can always change our mind later.

src/main/webapp/js/views/BaseView.js
mbucknell: I think this is fine to put here. If different views end up changing the route it's easier to do it here. Then the route fragments are only in two places (the AppRouter and here).

mbucknell:
@mamcderm-usgs , I just thought about this after answering some questions from Jessica. You will need to update the web.xml and add context variables to define the forward-url for the services that you are using. Right now they are hardcoded. We are now using Ansible to deploy configuration information for Enddat much like is done for NWC. See the repo: http://cida-eros-stash.er.usgs.gov:7990/stash/projects/EN/repos/enddat-configuration/browse . So you'll need to update the web.xml and then update the configuration repo with the context variables.

PR #21, opened 2016-03-11
Enddat 138

mamcderm-usgs:


mamcderm-usgs:
@mbucknell - If you have time can you comment on how I am doing the SiteModel promise.  I still need to do a fetch on the SiteModel (in updateMarker)  when the lat/lng changes but I can't seem to get the flow right with respect to the promise.  Or, preferably a call might be more effective to talk through the UI events, as of now, and for the next ticket.

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,178 @@
...
+
+						});
+						
+						self.set({sites: siteData});
+						log.debug('Fetched sites ' + _.size(siteData));
+						sitesDeferred.resolve();
+					},
+					error: function(jqXHR, textStatus, errorThrown) {
+						if (404 === jqXHR.status) {
+							log.debug('No NWIS data available: ' + textStatus);
```
mbucknell: You need to reject or resolve the sitesDeferred within this part of the if clause too.

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -22,7 +24,35 @@ define([
...
+		initialize: function (options) {			
+			/* 
+			 * Check to see if the url contains the parameters needed
+			 * to fetch data for the project.  Currently only checking
+			 * for the NWIS data type but will add others in the future.
+			 */
+			if (this.model.get('step') === this.model.CHOOSE_DATA_STEP &&
+				this.model.get('radius') &&
+				this.model.get('datasets') &&  //should we check for one and only one?
+				_.each(this.model.get('datasets'), function(el) {
```
mbucknell: It would be clearer if you use _.contains rather than _.each

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -31,11 +61,17 @@ define([
...
 
+			//pass in siteData to mapView where mapView listen to events
+			//on the siteData model. Specifically, there is a 'sync' event that is
+			//fired after successful fetching of the data.
 			this.mapView = new MapView({
 				el : this.$(MAPVIEW_SELECTOR),
 				mapDivId : 'map-div',
-				model : this.model
+				model : this.model,
+				siteModelPromise : this.siteModelPromise,
```
mbucknell: You should not be passing the promise into another view. The mapView should be handling events on the siteModel. So pass in the siteModel instead. This way the mapView can listen for sync or change events. Are you not getting the sync event? I suspect that because we are redefining fetch, we may need to trigger the sync event in the fetch call.

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -22,7 +24,35 @@ define([
...
+				this.siteModelPromise = this.siteModel.fetch(this.model);
+				this.siteModelPromise.fail(function() {
+					this.goHome();
+				});
+			/* Check if a siteModel was already created but clear the sites
+			 * attribute since the url does not have all the parameters to
+			 * fetch data.  This will preserve the parameter and statistic code
+			 * attributes in the siteModel.
+			 */
+			} else if (this.siteModel) {
```
mbucknell: I don't understand this part of the else if clause. The siteModel is only created in initialize so this clause will never be taken. 

Consider reorganizing this if clause to move the this.siteModel instantiation out of the if clause. Leave in the call to this.siteModel.fetch if the dataset is NWIS and the bounding box is defined. Then add model listeners (this.model) to fetch or clear the this.siteModel as appropriate.

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -22,7 +24,35 @@ define([
...
+				this.model.get('radius') &&
+				this.model.get('datasets') &&  //should we check for one and only one?
+				_.each(this.model.get('datasets'), function(el) {
+					if(el === DATATYPE_NWIS) {
+						return true;
+					}
+				})) {
+				
+				this.siteModel = new SiteModel();				
+				this.siteModelPromise = this.siteModel.fetch(this.model);
```
mbucknell: I think I mentioned this before, but consider just passing in the parts of the model that you need (in this case, the location and radius). I'm not a big fan of passing in the whole model.

src/main/webapp/js/views/MapView.js
```diff
@@ -152,6 +160,27 @@ define([
...
+		 *							latitude and longitude in order to be a valid location
+		 */
+		updateSiteMarker : function(promise) {
+			var self = this;
+			var siteIcon = mapUtils.createIcon("img/time-series.png");
+			promise.done(function() {
+				var sites = self.sites.get('sites');
+				//I think will need to remove existing sites if handling a mouse event
+				_.each(sites, function(el) {
+					var marker = new L.marker([el['lat'], el['lon']], {icon: siteIcon, title: el['name']}).addTo(self.map);
```
mbucknell: If you use L.marker (note that marker is lower case), you do not need the new. Seems like the above should do the trick.

PR #22, opened 2016-03-14
Enddat 138

mamcderm-usgs:


mamcderm-usgs:
@mbucknell - Seems to be working as we discussed.  I still have tests and the context variables/web.xml changes but thought I'd go ahead with fresh pull request if you have time to look.

src/main/webapp/js/views/MapView.js
```diff
@@ -152,6 +158,28 @@ define([
...
+
+		/*
+		 * Updates or adds a marker for each site
+		 * @param {WorkflowStateModel} model
+		 * @param {SiteModel} sites - has one or more site objects, each with properties
+		 *							latitude and longitude in order to be a valid location
+		 */
+		updateSiteMarker : function(sites) {
+			var self = this;
+			var siteObjects = this.sites.get('sites');
```
mamcderm-usgs: this.sites.get('sites'); seems to work properly but I don't understand why I can't reference "this.sites" as just "sites" from the passed in parameter, it gets sites is undefined in that case.  Why is that?

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,179 @@
...
+							});
+
+							siteData[el["site_no"]] = site;
+
+						});
+						
+						self.set({sites: siteData});
+						log.debug('Fetched sites ' + _.size(siteData));
+						sitesDeferred.resolve();
+						self.trigger('sync');
```
mbucknell: You need to pass into the trigger command the arguments that your sync event handler expects. According to the documentation that is the model and the response (so probably data). In your trigger above, if you are going to use the model in the sync event handler, you will need to add a second parameter so that it is self.trigger('sync', [self])

src/main/webapp/js/views/MapView.js
mbucknell: See comment in SiteModel.

src/main/webapp/js/models/SiteModel.js
mamcderm-usgs: Still having some trouble grasping what backbone is doing, but prior to the sync, the self.set({sites: siteData}) occurred so shouldn't the data be "in the model"?  It seems to work without the "data" parameter so that's why I'm asking but I'm probably missing something.

mbucknell: You are defining what gets sent to the event handler so put the model in the call to trigger than it won't be used when that event handler is called.. So yes, it will work without the data parameter because you are not using it in your sync event handler. I think it's fine to leave it out. This isn't a backbone issue. You might consider stepping into the "trigger" method to see exactly what is going on when you call trigger.

mbucknell:
I did a quick look. I'll do a thorough code review when you are ready unless you want me to do that now. You may also want to go back to the closed pull requests and make sure the comments have been addressed.

PR #25, opened 2016-03-21
Enddat 138

mamcderm-usgs:


src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,184 @@
...
+					var columns = {
+						"parameter_cd" : null,
+						"parameter_nm" : null
+					};
+
+					parsedParams = rdbUtils.parseRDB(lines, columns);
+					_.each(parsedParams, function(el, index) {
+						paramCodes[el["parameter_cd"]] = el["parameter_nm"];
+					});
+					self.set({parameterCodes: paramCodes});
```
mbucknell: Do we need to put the parameter codes and statistic codes in the model attributes? In other words, do these need to be visible outside the SiteModel module? If they are only used for parsing the nwis sites service, consider putting these as properties rather than assigning them to attributes on the model.

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,184 @@
...
+							"begin_date" : null,
+							"end_date" : null,
+							"count_nu" : null
+						};
+
+						parsedSites = rdbUtils.parseRDB(lines, importantColumns);
+						_.each(parsedSites, function(el, index) {
+							var site;
+							//Add the info to the sites
+							if (!_.has(siteData, (el["site_no"]))) {
```
mbucknell: So sometimes a site comes without a site_no? In that case, we just get a station name and lat/lon?

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,184 @@
...
+						
+						self.set({sites: siteData});
+						log.debug('Fetched sites ' + _.size(siteData));
+						sitesDeferred.resolve();
+						self.trigger('sync', self);
+					},
+					error: function(jqXHR, textStatus, errorThrown) {
+						if (404 === jqXHR.status) {
+							log.debug('No NWIS data available: ' + textStatus);
+							sitesDeferred.resolve();
```
mbucknell: So do we get 404 if there are no NWIS sites in our bounding box? If that is the case, then in addition to resolving sitesDeferred, we should also clear the sites from the model and trigger the 'sync' event. If its an error condition we should not be resolving the deferred.

src/main/webapp/js/models/SiteModel.js
```diff
@@ -0,0 +1,184 @@
...
+	'module',
+	'utils/rdbUtils',
+	'utils/geoSpatialUtils'
+], function ($, Backbone, log, module, rdbUtils, geoSpatialUtils) {
+	"use strict";
+
+	var config = module.config();
+	var pcodes = config.parameterCodesPath;
+	
+	var model = Backbone.Model.extend({
```
mbucknell: As we move forward, we may want to reconsider using a Model rather than a Collection. This really represents a set of NWIS sites, not a single site as the module name implies. I don't think it needs to be changed now, but something to consider for the future.

src/main/webapp/js/utils/rdbUtils.js
```diff
@@ -0,0 +1,58 @@
...
+					});
+
+					columnsArray.push(columnValues);
+				}
+			}
+		});
+		return columnsArray;
+	};
+
+	self.toTitleCase = function (str) {
```
mbucknell: Add a comment so that we know what this function does.

src/main/webapp/js/views/BaseView.js
```diff
@@ -50,6 +50,11 @@ define([
 			}
 
 			Backbone.View.prototype.initialize.apply(this, arguments);
+		},
+
+		/* COMMON navigation functions that might be used for many views */
+		goHome: function(options) {
+			this.router.navigate();
```
mbucknell: Shouldn't this be this.router.navigate('', options)?

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -60,6 +70,26 @@ define([
...
+
+		/* 
+		 * Check to see if the WorkflowStateModel contains the parameters needed
+		 * to fetch data for the project.  Currently only checking
+		 * for the NWIS data type but will add others in the future.
+		 */
+		updateSiteModel: function () {
+			self = this;
+			if (this.model.get('step') === this.model.CHOOSE_DATA_STEP &&
+				this.model.get('radius') &&
```
mbucknell: You should also check that the 'location' property is defined before fetching and that it has both a lat and lon property. You will probably also want to make sure the radius property is non null.

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -60,6 +70,26 @@ define([
...
+				_.contains(this.model.get('datasets'), DATATYPE_NWIS)) {
+
+				this.siteModel.fetch(this.model.get('location'), this.model.get('radius')).fail(function() {
+					self.goHome();
```
mbucknell: You don't need to change this now, but as you move forward, we should not be routing back home. Rather we will need to add some sort of loading indicator and an alert area where we can tell the user that data fetching has failed. You can add that to the next ticket.

src/main/webapp/js/models/SiteModel.js
mamcderm-usgs: I wasn't sure if we needed them somewhere later but probably not.  I'll go ahead and put them in properties.

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -60,6 +70,26 @@ define([
...
+				});
+			}
+			else {
+				this.siteModel.clear();
```
mbucknell: The way siteModel is currently implemented, you would not use clear because that would clear the parameter codes and site codes ... another good reason to move these from model attributes to properties on the model instance. If you do use clear you will need to add event handlers for the change event in MapView so that you can clear the NWIS site markers.

src/main/webapp/js/views/MapView.js
```diff
@@ -152,10 +158,32 @@ define([
...
+				this.map.removeLayer(this.siteLayerGroup);
+			}
+			this.siteLayerGroup = L.layerGroup();
+			if(!_.isEmpty(siteObjects)) {
+				_.each(siteObjects, function(el) {
+					var marker = L.marker([el['lat'], el['lon']], {icon: siteIcon, title: el['name']});
+					self.siteLayerGroup.addLayer(marker);
+				});
+				this.siteLayerGroup.addTo(this.map);				
+			}
```
mbucknell: This code works the first time, but on subsequent calls it won't work. There will already be a siteLayerGroup the second time you come into this function. So you can handle this several ways, but that group either needs to be removed from the map and recreated or the markers should be removed from the group prior to creating new ones. You should be able to test this by changing the location in the UI.

src/test/js/views/DataDiscoveryViewSpec.js
```diff
@@ -27,6 +29,18 @@ define([
 			$('body').append('<div id="test-div"></div>');
 			$testDiv = $('#test-div');
 
+			initializeSiteModelSpy = jasmine.createSpy('initializeSiteModelSpy');
+			fetchSiteModelDoneSpy = jasmine.createSpy('fetchSiteModelDoneSpy').and.callFake(function(){
+				var d = $.Deferred();
+				d.resolve();
+				return d.promise();
+			});
+			fetchSiteModelFailSpy = jasmine.createSpy('fetchSiteModelFailSpy').and.callFake(function(){
```
mbucknell: I don't think this is used so it should be removed. 

I have been using a pattern where I create a global opDeferred Deferred object, which will be returned from my spied fetch function. Then within a specific test, I can resolve or reject the deferred to test the success or failure of the fetch. See https://github.com/USGS-CIDA/PubsWarehouse_UI/blob/master/tests/js/manager/views/PublicationViewSpec.js for an example.

src/test/js/views/MapViewSpec.js
```diff
@@ -98,6 +103,12 @@ define([
 				expect(removeMapSpy).toHaveBeenCalled();
 				expect(L.map.calls.count()).toBe(2);
 			});
+
+			it('Expect that the map location marker is added to the map if there are sites in the site model', function() {
```
mbucknell: I don't understand this test. The map location marker gets added to the map whether there are sites in the site model or not. If you are trying to test if site markers are added, you are going to need to test this differently.

src/test/js/views/MapViewSpec.js
```diff
@@ -137,12 +148,18 @@ define([
...
 				hasLayerSpy.and.returnValue(true);
 				expect(removeLayerSpy).not.toHaveBeenCalled();
 				testModel.set('location', {});
 				expect(removeLayerSpy).toHaveBeenCalled();
 			});
+
+			it('Expects that if the site model is updated, an updated site marker is added to the map', function() {
+				testSiteModel.set({'sites': {'05464220': {'name': 'test', 'lat': '42.25152778', 'lon': '-92.2988889'}}});
+				testSiteModel.trigger('sync', testSiteModel);
+				expect(addLayerSpy).toHaveBeenCalled();
```
mbucknell: I think you need to be careful using these spies (addLayerSpy). They are called prior to anything happening with the event handler. You either need to clear the spy (see http://jasmine.github.io/2.4/introduction.html#section-35) or see how many times they have been called prior to your triggering event and then comparing the number of times the spy has been called. And they are not called to add site markers as this is not done with addLayer

src/main/webapp/js/views/MapView.js
mamcderm-usgs: This does seem to work in the UI.  I thought I was handling this with line 174.  Does that not remove the group on subsequent calls?

mbucknell: You are correct. That will remove it. Sorry!

src/test/js/views/DataDiscoveryViewSpec.js
mamcderm-usgs: That pattern is excellent.  Unfortunately, I cannot figure out how to apply it when the model is instantiated in the view.  I'm sure I must be missing something simple.  I've spent so much time on this I wish I could figure it out but I'll just cleanup the spy and leave it for now.

mbucknell: I think all you have to do is move the var d outside of the returned function so that it is visible in the tests. I may be missing something though. The other option would be to do refactor and pass in the empty SiteModel into the DataDiscoveryView. But no need to do anything now other than clean up the unused variable. I have opened three tickets to improve test coverage, including one for this module.

src/test/js/views/MapViewSpec.js
mamcderm-usgs: On line 180 in MapView.js:
180					self.siteLayerGroup.addLayer(marker);
I thought that was using addLayer to add the site markers.  Should I not be doing it that way?

src/test/js/views/DataDiscoveryViewSpec.js
mamcderm-usgs: I would have thought so too!  I tried that and several other variations but it just didn't seem to work, but I am probably doing something stupid.

src/test/js/views/MapViewSpec.js
mbucknell: It's fine to do it that way in MapView. But your test is looking at the addLayerSpy which is attached to the map object. You are calling addLayer on siteLayerGroup and I don't think you have a spy for that at all.

mamcderm-usgs: Okay, I see what you're saying, I should be more explicit and setup a spy for siteLayerGroup.  But, the test does seem to work when I clear the addLayerSpy and check if it was called, so the siteLayerGroup.addLayer() must also hit that spy?  Or is some other event hitting that spy inadvertently.

mbucknell: Probably because it is called within siteLayerGroup to add the layer. The problem is you don't really know from this test that it added your siteLayerGroup unless you examine the arguments.I'm fine with leaving the test as is for now since you are now clearing the addLayerSpy.

PR #32, opened 2016-03-25
Enddat 150

mbucknell:
Added an opened option to BaseCollapsibleView so that when these views are instantiated the view can be rendered with the panel open or collapsed.

To make the cursor the hand pointer when first clicking on the map, you must set the .leaflet-clickable class on the tile layers and then remove it when no longer desired.

Also added the ability to update the extent when the location is changed.

### DIFF
--- a/src/main/webapp/js/views/DataDiscoveryView.js
+++ b/src/main/webapp/js/views/DataDiscoveryView.js
@@ -1,13 +1,14 @@
 /* jslint browser: true */
 
 define([
+	'loglevel',
 	'views/BaseView',
 	'views/NavView',
 	'views/MapView',
 	'views/LocationView',
 	'models/SiteModel',
 	'hbs!hb_templates/dataDiscovery'
-], function (BaseView, NavView, MapView, LocationView, SiteModel, hbTemplate) {
+], function (log, BaseView, NavView, MapView, LocationView, SiteModel, hbTemplate) {
 	"use strict";
 
 	var NAVVIEW_SELECTOR = '.workflow-nav';
--- a/src/main/webapp/js/views/DataDiscoveryView.js
+++ b/src/main/webapp/js/views/DataDiscoveryView.js
@@ -89,7 +89,7 @@ define([
 				_.contains(this.model.get('datasets'), DATATYPE_NWIS)) {
 
 				this.siteModel.fetch(this.model.get('location'), this.model.get('radius')).fail(function() {
-					self.goHome();
+					log.debug('Fetch failed');
 				});
 			}
 			else {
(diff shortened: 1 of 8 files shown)


src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -88,7 +90,7 @@ define([
 				_.contains(this.model.get('datasets'), DATATYPE_NWIS)) {
 
 				this.siteModel.fetch(this.model.get('location'), this.model.get('radius')).fail(function() {
-					self.goHome();
+					log.debug('Fetch failed');
```
mbucknell: @mamcderm-usgs, I assume that you will be working on providing better error handling as well as providing a loading indicator. Routing to home is not what we want to do if there is an error and made it difficult to test. For now, I just put in a log message. Also if a 404 is returned, that is not an error but means that no sites were found so the model should be cleared. This should be done in the SiteModel module.

mamcderm-usgs:
Looks great :+1:

src/main/webapp/js/views/DataDiscoveryView.js
mamcderm-usgs: I was planning to put in a loading indicator, I'm thinking it can just "wrap" the fetch.  I'm not certain how to best handle errors.  Currently, I'm handling an error and a 404 in the site model in the same way, clear the model, trigger sync and reject the deferred.    Both of these cases are handled with the fail deferred method in DataDiscovery.js.  I think you are saying that is not correct.  Do you have any suggestions for how to handle the two ajax error condtions?  Should we be using some sort of warning dialogue to notify the user?  I had brought the showWarningDialog from the original app but you suggested to not include that...was that just at that time or did you have a different preference for user notification?

mbucknell: I think a loading indicator directly above the map would work. Next to it could go a bootstrap alert, http://getbootstrap.com/javascript/#alerts. We used these in pubs, so we may want to do something similar. There is an AlertView in pubs that we might be able to use: https://github.com/USGS-CIDA/PubsWarehouse_UI/blob/master/pubs_ui/manager/static/js/views/AlertView.js.

I'm currently working on loading precipitation data and will need a loading indicator and error area too. Why don't I add that and then you should be able to reuse.

In regards to handling 404 versus other errors. I think these should be handled differently. 404 is a normal response that occurs when no sites are within the bounding box, therefore the model should be cleared and the deferred resolved. For other errors, we should also probably clear the model, but the deferred should be rejected.

mamcderm-usgs: That would be great if I could reuse the loading indicator and error area, thank you.  I agree about the 404 versus other errors and I'll make that change.

PR #33, opened 2016-03-28
Enddat 139

mamcderm-usgs:
@mbucknell  - I think this is in a pretty good state if you want to add the dataset multi picker to the Choose Data panel sooner than later (review/merge or however you want to proceed).  The radius text box is working properly but I haven't added begin and end date yet.  Also, haven't added the load spinner and error indicator (for nwis data).  I can keep working on this to add the dates and some additional testing, so no problem if this is not a good time or "point in the code" to try and sync.

### DIFF
--- a/src/main/webapp/js/controller/AppRouter.js
+++ b/src/main/webapp/js/controller/AppRouter.js
@@ -5,8 +5,9 @@ define([
 	'backbone',
 	'loglevel',
 	'models/WorkflowStateModel',
+	'models/SiteModel',
 	'views/DataDiscoveryView'
-], function ($, Backbone, log, WorkflowStateModel, DataDiscoveryView) {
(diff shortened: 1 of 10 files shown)


mbucknell:
I had forgotten about the temporal filter. I don't need to add the picker for the data set in order to test what I am doing. I have already added the loading indicator and alert area. I'm starting work on adding tests. So I may be able to hand this back to you to add the multi select picker for the datasets by tomorrow.

mamcderm-usgs:
Okay, no rush I think I may have misunderstood Jessica.  We talked on Friday and I thought I was just adding the radius for now to keep from holding you up.  I thought she indicated you were going to add the dataset multipicker but it sounds now like you are just doing the "model side" of that.  I can go ahead and add the date fields and improve testing but I will wait to follow your work on the loading indicator and alert area.

mbucknell:
The problem isn't so much that you would hold me up, but that we would be working on the same pieces of the code.

PR #34, opened 2016-03-29
Enddat 151

mbucknell:
@mamcderm-usgs , I'm not finished yet. The code all works and the tests run, but I still need to add tests for the WorkflowModel changes and possibly some other utilities I added. The biggest changes I made are in WorkflowModel which now contains a datasetModels property which is an array of dataset model objects. Right now there is the SiteModel and PrecipitationCollection. Added listeners for the location, radius changes to here from DataDiscoveryView because it seemed to make more sense. It keeps the DataDiscoveryView much cleaner. I'm creating my own events in WorkflowModel to signal when the dataset model fetches start and when they are complete

I can merge this if you think it will be helpful.


mamcderm-usgs:
I think it would be best if you go ahead and merge.

mbucknell:
ok , I'll go ahead and merge.

PR #35, opened 2016-03-30
Enddat 139

mamcderm-usgs:


src/main/webapp/js/views/ChooseView.js
```diff
@@ -0,0 +1,55 @@
...
+	 */
+	var view = BaseCollapsiblePanelView.extend({
+		template : hbTemplate,
+
+		panelHeading : 'Choose Data',
+		panelBodyId : 'choose-data-panel-body',
+
+		bindings : {
+			'#radius' : {
+				observe : 'radius',
```
mbucknell: You don't need to specify onGet and onSet separately. You can just do '#radius' : 'radius'

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -106,7 +118,22 @@ define([
 			else {
 				this.alertView.showDangerAlert('Unable to fetch the following data types: ' + fetchErrorTypes.join(', '));
 			}
+			this.$(ALERTVIEW_SELECTOR).show();
+		},
+
+		updateChooseView: function () {
```
mbucknell: This is something to consider but you don't need to change now. I was thinking that we would create/render the and remove the views as needed, rather than creating all of them at initialization and then hiding there divs. This is find to leave as is. However, I don't think the alertview should ever be hidden. It should always be available for alert messages.

mbucknell:
A couple of comments that you can deal with later if you want. You can go ahead and merge.

src/main/webapp/js/views/DataDiscoveryView.js
mamcderm-usgs: If Specify Project Location is clicked after Choose Data, do you prefer some informational message to display or to "clear" the alertview?  If "clear" is best, is a "clearAlert" function okay which passes a blank "kind" and "message" into showAlert, or is there a better practice to use for clear.

mbucknell: Let's not worry about this, here. In talking with Dave and Jessica, we are going to do some "Are you sure" modals when switching navigation steps.

mbucknell: We may want to eventually implement a closeAlertView function in AlertView to remove the area. But the user can always dismiss the area.

mamcderm-usgs: Okay, well I left the hide() in for now to hide the message you get on Choose Data, but then click on Specify Location.  Let me know if you want me to implement the closeAlertView as I continue to work on Choose Data.  I would think we should do the same with the Choose Data view, rather than show/hide as you indicated.

PR #36, opened 2016-03-30
Enddat 151

mbucknell:
The major changes between this and the previous pull request was added tests for the new functionality. Also changed to add a call to explictly initialize the data set models in workflowStateModel. This way any info such as the parameter codes does not get loaded until its needed.


mamcderm-usgs:
Wunderbar :+1:

mbucknell:
I'm going to close this one. The refactoring I'm doing is significant enough that I don't want to just add it to this pull request.

PR #38, opened 2016-03-31
Enddat 151

mbucknell:
Added tests for PrecipitationCollection, AlertView, and jqueryUtils. Refactored SiteModel to be a collection (SiteCollection) and update usage. Decided to move into WorkflowStateModel the code to update the model attributes when the step changes. Now NavView is only responsible for updating the step and then handling the model change event. I suspect this code is still not quite right, but adding tests so that when we encounter a new situation we can add a test before implementing a fix.


mbucknell:
@mamcderm-usgs , This got to be bigger than I anticipated. The refactoring of SiteCollection was straightforward but then I needed to fix the navigation between steps. I also did add a closeAlert to AlertView as well as change DataDiscoveryView to create/render the sub views when needed and then removing them when they are not needed.

src/test/js/views/NavViewSpec.js
```diff
@@ -7,7 +7,7 @@ define([
 	'views/BaseView',
 	'views/NavView'
 ], function($, log, WorkflowStateModel, BaseView, NavView) {
-	describe('views/NavView', function() {
+	fdescribe('views/NavView', function() {
```
mamcderm-usgs: typo?

mbucknell: Fixed that locally and forgot to push it up. the fdescribe is for focused spec. You can use that to run that particular suite of tests.

mamcderm-usgs:
Wow, that is incredible, quite a refactor.  It's impressive to see you iteratively progress through building this app.  I feel good that I can follow what you're doing but I have a long way to go to be able to do this.  I think I get stuck trying to understand how it's all suppose to work instead of just write/refactor.  Nice job :+1:

PR #39, opened 2016-04-04
Enddat 139

mamcderm-usgs:


bower.json
```diff
@@ -23,7 +23,9 @@
     "handlebars": "~4.0.3",
     "Squire.js": "https://github.com/iammerrick/Squire.js.git#0.2.0",
     "font-awesome": "fontawesome#~4.4.0",
+    "select2": "~4.0.0",
```
mbucknell: Latest version of this is 4.0.2. Is there a reason we aren't using the latest?

src/main/webapp/js/models/WorkflowStateModel.js
```diff
@@ -122,6 +123,8 @@ define([
 				_.each(datasetCollections, function(datasetCollection) {
 					datasetCollection.reset();
 				});
+				//if clearing datasets through UI, this will clear alert from previous fetch
+				self.trigger('dataset:updateFinished', undefined);
```
mbucknell: I'm not a big fan of triggering this event in this situation. I envisioned that updateStarted and updateFinished were a pair that get triggered when dataset fetching occurs.

Consider putting an event handler in DataDiscoveryView for a change on the dataset attribute that clears the alert whenever the dataset attribute is cleared. Not sure this will give us the behavior that we want so feel free to leave as is for now.

mamcderm-usgs: I didn't like this either since it kind of hijacked the start/finish pair.  I think a handler in the DataDiscoveryView to clear the alert should work, but can we get away with not resetting the datasetCollection?  Or, should it be reset by calling updateDatasetCollections() from the DataDiscoveryView event handler, or maybe there is another obviously simple way I am missing.

bower.json
mamcderm-usgs: No reason, I meant to change that and forgot.  I'll make sure to change it.

src/main/webapp/js/models/WorkflowStateModel.js
mbucknell: When datasets attribute is updated in ChooseView, it triggers the event handler in WorkflowStateModel to updateDatasetCollections. So that's already handled. DataDiscoveryView would have an event handler 'change:datasets' that would clear the alert. So DataDiscoveryView should not have to call updateDatasetCollections.

src/main/webapp/js/views/ChooseView.js
```diff
@@ -22,7 +24,28 @@ define([
 		panelBodyId : 'choose-data-panel-body',
 
 		bindings : {
-			'#radius' : 'radius'
+			'#radius' : 'radius',
+			'#datasets-select' : {
+				observe : 'datasets',
+				update : function($el, value, model) {
```
mbucknell: The way I am reading the backbone stickit documentation,the update gets called whenever the model value changes. It doesn't seem like we should be initializing the select2 each time this happens. I'm worried that repeated initialization without destroying the select2 may cause memory leaks or other bad behavior. Can you verify that this only gets called once by setting breakpoints in the function and then updating the chosen datasets? Would it be more appropriate to use initialize rather than update?

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -142,7 +142,10 @@ define([
 			var chosenDatasets = this.model.get('datasets');
 
 			this.$(LOADING_SELECTOR).hide();
-			if (fetchErrorTypes.length === 0) {
+			if (fetchErrorTypes === undefined) {
```
mbucknell: You won't need this if you don't trigger the updateFinished event when clearing the dataset collections.

src/test/js/views/DataDiscoveryViewSpec.js
```diff
@@ -26,6 +26,7 @@ define([
 		var injector;
 
 		beforeEach(function(done) {
+			sinon.stub($, "ajax");
```
mbucknell: Is this needed? By the way, I'm assuming you are using this so Squire can still fetch the required js files while still stubbing out the ajax calls. Great idea!

src/main/webapp/js/models/WorkflowStateModel.js
mamcderm-usgs: Got it...was thinking you meant to also remove the "this.on datasets", but you only said the trigger.  I put the event handler in the DataDiscoveryView and works great.

src/main/webapp/js/views/DataDiscoveryView.js
mamcderm-usgs: Yep, cleaned this up.

src/test/js/views/DataDiscoveryViewSpec.js
mamcderm-usgs: Wish I could claim I did something smart here but can't really make that case.  I put that in there to prevent getting 404 errors due to the model instantiations that were getting triggered from a "step change" (line 33 of WorkflowStateModel) which calls updateModelState which calls intializeDatasetCollections.  I couldn't see how to mock it out so I just mocked the ajax call.  Not sure if that's the best way to handle the situation....

src/main/webapp/js/views/ChooseView.js
mamcderm-usgs: I don't think it's getting updated whenever the model changes, at least it doesn't stop in the update function every time the dataset changes in the UI, only on the "initial binding".  I tried just replacing update with initialize and that didn't work, so I'm not sure if there's more to it, or if initialize just isn't the right approach here.  I'm going to investigate some more after lunch but honestly I'm not sure that I know enough to confidently address your concern.

mbucknell: I looked back at the pubs code to see why I did what I did. The comment in that code tells me that the binding from the model to the DOM was not working when using stickit. What I'm afraid of is that your solution works because right now we are really only doing one way binding ... from the DOM to the model except at page load.  I think its fine to leave as is if you want... but we need to be aware going forward.

mamcderm-usgs: It seems like it works to me and I thought the tests proved that a change to the model will cause a change to the DOM, through the binding?  Also, this sentence in under the "Form Element Bindings" section of the stickit documentation made me think it should work:  
By default, form and contenteditable elements will be configured with two-way bindings, syncing changes in the view elements with model attributes.
But I could be misunderstanding exactly what is meant by "binding" and it's not really doing everything we need it to do.  Nevertheless, I'm going to leave it for now since I'm not certain how to thoroughly evaluate it.

mbucknell: That sentence refers to input types as well as textarea and selects. This is a select, however when you make it a select2 other DOM elements are added so that sentence doesn't necessarily apply. I think its fine to leave as is. As I said, we really aren't using two way binding at the moment.

mbucknell: And just a note about the tests. As a said before its not really a select, so you don't know if visually the selection appears as you have set it. There are things we just can't test or that are very difficult to test.

PR #40, opened 2016-04-07
Enddat 152

mbucknell:
Added the ability to click on a precipitation grid point. This creates a circle marker at that point and shows the PrecipDataView which allows the user to examine the data and select it for processing. The select state is saved within the precipitation data model.


src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -87,20 +88,21 @@ define([
 				case model.PROJ_LOC_STEP:
 					if (!this.locationView) {
 						this.locationView = new LocationView({
-							el : this.$(LOCATION_SELECTOR),
+							el : $utils.createDivInContainer(this.$(LOCATION_SELECTOR)),
```
mbucknell: We need to create the div to put the sub views in so that when they are removed it doesn't remove the container that is part of the dataDiscovery.hbs template.

src/main/webapp/js/hb_templates/precipData.hbs
```diff
@@ -0,0 +1,24 @@
+<div class="precip-data-select">
+	<table class="table">
+		<thead>
+			<tr>
+				<th></th>
+				<th>y:x</th>
```
instantsoup: Y:X instead of X:Y? Is this a crazy map thing?

src/main/webapp/js/models/WorkflowStateModel.js
```diff
@@ -59,15 +59,21 @@ define([
 			}
 		},
 
+		hasValidLocation : function() {
+			return (this.has('location') &&
+				((this.attributes.location.latitude) ? true : false) &&
```
instantsoup: if this.attributes.location.latitude resolves to true or false, you don't need a ternary op

src/main/webapp/js/hb_templates/precipData.hbs
mbucknell: It's actually y:x on the old enddat, just labeled as x:y. It corresponds to lat:lon

src/main/webapp/js/models/WorkflowStateModel.js
mbucknell: It doesn't ... it's a string.

src/main/webapp/js/views/MapView.js
```diff
@@ -209,15 +252,56 @@ define([
...
+			var moveCircleMarker = function(latLng) {
+				if (self.circleMarker) {
+					self.circleMarker.setLatLng(latLng);
+					self.map.addLayer(self.circleMarker);
+				}
+				else {
+					self.circleMarker = new L.circleMarker(latLng,{
+						radius : 15
+					});
+					if (!self.map.hasLayer(self.circleMarker)) {
```
instantsoup: Checking map for layer but didn't check 7 lines above. Does it matter? Can you extract the addLayer call to outside the conditional if doing it for both cases anyway?

mbucknell: I think I have this logic backwards anyways. If the circleMarker exists, all I have to do is update its lat lng. If it doesn't exist, then just create it and add to map. I'll fix this.

src/main/webapp/js/models/WorkflowStateModel.js
```diff
@@ -59,15 +59,21 @@ define([
...
 		/*
 		 * Returns the bounding box as an object with west, east, north, and south properties.
 		 * Return undefined if the model's properties do not contain a valid bounding box
 		 * @return {Object} - has west, east, north, south properties or undefined
 		 */
 		getBoundingBox : function() {
 			var result = undefined;
-			if ((this.attributes.radius) && (this.attributes.location) &&
-				(this.attributes.location.latitude) && (this.attributes.location.longitude)) {
+			if ((this.attributes.radius) && this.hasValidLocation()) {
```
instantsoup: Is radius a boolean? I believe you set it to 15 creating a new one. It's a nitpick, but there's ternary usage and then this.

mbucknell: It isn't a boolean, but isn't this usage fine since it is returning a truthy/falsy value?

instantsoup: Yeah, but how's it different than the comparison in the hasValidLocation() method? Why is that one ternary?

mbucknell: I did have it like that, but when I wrote the tests for hasValidLocation, it was returning the longitude value if it was a location with both a latitude and longitude. I was surprised it did that, but I think in that case, I need to return a boolean value, not something that is "truthy".

PR #48, opened 2016-05-03
Enddat 172

mbucknell:
The project was already being compiled with java 8. I wanted to update the proxy-utils to the latest version and try to eliminate use of the AuthIgnoringProxyServlet that we were using to read the NWIS statistic codes service. I was able to do this by modifying the calling parameters slightly and by updating the rdb parsing utility to handle lines that ended in a carriage return and line feed. 

As part of this work I refactored the rdp parsing utility to take the entire file contents as a string, eliminating duplicative code in NWISCollection and by removing the importantColumns parameters, leaving it to the caller of the parsing to modify further than returned array.

### DIFF
--- a/src/main/webapp/js/utils/rdbUtils.js
+++ b/src/main/webapp/js/utils/rdbUtils.js
@@ -4,47 +4,40 @@ define([
 	'underscore'
 ], function (_) {
 	"use strict";
-	
+
 	var self = {};
 
-	/* @param	[Array] lines - array containing each line from an rdb file
-	 * @param	{Object} importantColumns - object of tab-delimited columns found in lines
-	 * returns	[Array] columnsArray - array of objects where each object holds key/value(s)
-	 *					corresponding to the parsed column(s) from a line
+	/*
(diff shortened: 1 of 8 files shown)


src/main/webapp/js/utils/rdbUtils.js
```diff
@@ -4,58 +4,52 @@ define([
...
 
-	/* @param	[Array] lines - array containing each line from an rdb file
-	 * @param	{Object} importantColumns - object of tab-delimited columns found in lines
-	 * returns	[Array] columnsArray - array of objects where each object holds key/value(s)
-	 *					corresponding to the parsed column(s) from a line
+	/*
+	 * Function assumes that the file uses rdb formatting, that columns are separated by tabs, that rows
+	 * that begin with a # characters are comments, that the first non comment line are the column keys, and
+	 * that the line following this line does not contain data.
+	 * Note that the function handles Strings where the lines end in \n (linefeed) or in \r followed by \l
```
dsteinich: Should the last thing on this line be \n rather than \l

mbucknell:
Yes. I'll fix the comment.

PR #56, opened 2016-05-05
Enddat 173

mbucknell:
I added additional text to the successfully loaded message to indicate what filter was used to load and display the data. Since we want this message to be updated when the date filter changes, tied event handlers to the date filter attributes which show the alert message. Also modified the dataset:updateStarted and dataset:updateFinished signals to be sent, even when no datasets are loaded but only cleared.

### DIFF
--- a/src/main/webapp/js/views/ChooseView.js
+++ b/src/main/webapp/js/views/ChooseView.js
@@ -35,7 +35,10 @@ define([
 		},
 
 		bindings : {
-			'#radius' : 'radius'
+			'#radius' : {
+				observe : 'radius',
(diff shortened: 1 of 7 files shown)


src/main/webapp/js/views/ChooseView.js
```diff
@@ -35,7 +35,10 @@ define([
 		},
 
 		bindings : {
-			'#radius' : 'radius'
+			'#radius' : {
```
mbucknell: Made this change so that the radius property is only updated when tabbing out of the field or hitting enter, not on any text change.

PR #71, opened 2016-06-03
Enddat 188

mbucknell:
Added the ACIS service to the application. The service responds with JSON and all information needed to make the processing calls are included in the response. Was able to extend the standard Backbone fetch rather than redefining the whole method for the ACISCollection.

Sites show up on the map, but selected an ACIS site does nothing. Implementation of the next steps are part of ENDDAT-189.


ayan-usgs:
Might take this opportunity to poke at the Backbone documentation and tutorials so I can figure out what's going on.

mbucknell:
I'm going to merge this so I can make the next pull request.

PR #72, opened 2016-06-06
Enddat 189

mbucknell:
Refactored code by creating a BaseDataView which will be extended for each site collection. When extending this view, the template, panelHeading, panelBodyId, and render function will be redefined for the specific dataset. Extended this view to implement the NWISDataView, PrecipDataView and the new ACISDataView. 

Added the ACIS collection to VariableSummaryView and corrected the creation of the variableParameter for ACIS so that the process data feature works.

### DIFF
--- a/src/main/webapp/js/models/ACISCollection.js
+++ b/src/main/webapp/js/models/ACISCollection.js
@@ -34,7 +34,8 @@ define([
 
 		parse : function(response) {
 			var sites = response.meta;
-			var result = _.map(sites, function(site) {
+			return _.map(sites, function(site) {
+				var sid = site.sids[0].split(' ')[0];
 				var variables = _.chain(site.valid_daterange)
 					.map(function(dateRange, varIndex) {
 						var result = _.clone(ELEMS[varIndex]);
@@ -43,7 +44,7 @@ define([
 							result.endDate = moment(dateRange[1], Config.DATE_FORMAT);
 							result.variableParameter = new VariableParameter({
 								name : 'ACIS',
-								value : site.sids[0] + ':' +  result.code,
+								value : sid + ':' +  result.code,
 								colName : result.description
 							});
 						}
@@ -58,12 +59,10 @@ define([
 					lon : site.ll[0],
 					lat : site.ll[1],
 					name : site.name,
-					sid : site.sids[0],
+					sid : sid,
 					variables : new BaseVariableCollection(variables)
 				};
 			});
-
-			return result;
 		},
 
 		fetch : function(boundingBox) {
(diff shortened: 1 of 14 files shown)


src/main/webapp/js/models/ACISCollection.js
```diff
@@ -33,7 +34,8 @@ define([
 
 		parse : function(response) {
 			var sites = response.meta;
-			var result = _.map(sites, function(site) {
+			return _.map(sites, function(site) {
+				var sid = site.sids[0].split(' ')[0];
```
mbucknell: The enddat-services work only if the sid is sent, so don't send the sid type.

src/test/js/views/NWISDataViewSpec.js
```diff
@@ -89,41 +89,5 @@ define([
 
 			expect(testView.context.variables[0].selected).toBe(true);
 		});
-
```
mbucknell: This code is part of BaseDataView and is tested in that spec.

src/test/js/views/PrecipDataViewSpec.js
```diff
@@ -67,36 +67,5 @@ define([
 
 			expect(testView.$('input:checkbox').is(':checked')).toBe(true);
 		});
-
-		it('Expects that if the checkbox is checked, the selected property is set to true', function() {
```
mbucknell: The code being tested is part of BaseDataView and is tested in that spec.

src/main/webapp/js/views/ACISDataView.js
```diff
@@ -0,0 +1,46 @@
+/* jslint browser: true */
+
+define([
+	'underscore',
+	'Config',
+	'views/BaseDataView',
+	'hbs!hb_templates/acisData'
+], function(_, Config, BaseDataView, hbTemplate) {
+
+	"use strict";
```
ayan-usgs: What is the benefit `use strict`?

http://stackoverflow.com/questions/1335851/what-does-use-strict-do-in-javascript-and-what-is-the-reasoning-behind-it makes it sound like a few useful things, but it's unclear to me what exactly it's doing. It here to throw errors?

ayan-usgs:
Looks good to go. :+1:

src/main/webapp/js/views/ACISDataView.js
thongsav-usgs: Sorry, eaves dropping here (and I wish I used "use strict" more often), but the biggest reason I would encourage the use of it is that it catches when you forget to declare a variable with "var". In that situation, a global variable is created and used and that's prone to side affects (other parts of the code using the same variable).

That is mentioned in one of the posts further down (access to the global object).

mbucknell: Exactly why I use it. It catches common errors. My PyCharm IDE will actually add it for you.

PR #76, opened 2016-06-07
ENDDAT-175

instantsoup:
First blush for getting points on a map for GLCFS.


src/main/webapp/WEB-INF/jsp/head.jsp
```diff
@@ -23,6 +23,12 @@
 	String parameterCodesUrl = props.getProperty("enddat.endpoint.nwis.pmcodes");
 	String precipWFSGetFeatureUrl = props.getProperty("enddat.endpoint.precip.wfsgetfeature");
 	String cidaThreddsPrecipData = props.getProperty("enddat.cidathredds.precipdata");
+	String GLCFSWFSGetFeatureUrlErie = props.getProperty("enddat.endpoint.glcfs.wfsgetfeature.erie");
```
mbucknell: Just for consistency, can we make these glcfWFS...

src/main/webapp/js/models/GLCFSCollection.js
```diff
@@ -0,0 +1,160 @@
...
+	'underscore',
+	'moment',
+	'utils/VariableParameter',
+	'models/BaseDatasetCollection',
+	'models/BaseVariableCollection',
+	'utils/jqueryUtils'
+], function(log, module, $, _, moment, VariableParameter, BaseDatasetCollection, BaseVariableCollection, $utils) {
+	"use strict";
+
+	var startDate = moment('2016-01-01', 'YYYY-MM-DD');
```
mbucknell: This should be the start date for the data set which I'm pretty sure isn't Jan 1, 2016. Also (and I know I'm not consistent about this either), constants should be all uppercase.

src/main/webapp/js/models/GLCFSCollection.js
```diff
@@ -0,0 +1,160 @@
...
+	
+	var variableId = 'glcfstempVariableId';
+	var variableName = 'GLCFStemp';
+
+	var getInteger = function(str) {
+		return str.split('.')[0];
+	};
+
+	var getDDSURL = function(lake) {
+		var url = 'glosthredds/' + module.config().glosThreddsGLCFSData + lake.toLowerCase() + '/fcfmrc-2d/Lake_' + lake + '_-_2D_best.ncd.dds'
```
mbucknell: Do you think we should have thredds data context variable for each lake rather than forming the data part of the URL here? Just something to think about.

src/main/webapp/js/models/GLCFSCollection.js
```diff
@@ -0,0 +1,160 @@
...
+			$.ajax({
+				url : GLCFSWFSGetFeatureUrls[this.lake],
+				data : {
+					typeName : 'sb:' + this.lake.toLowerCase(),
+					srsName : 'EPSG:4269',
+					bbox : [boundingBox.south, boundingBox.west, boundingBox.north, boundingBox.east].join(',')
+				},
+				success : function(xml, textStatus, jqXHR) {
+					if ($utils.xmlFind($(xml), 'ows', 'ExceptionReport').length > 0) {
+						log.debug('CLCFS GetFeature fetch failed with Exception from service');
```
mbucknell: Typo - 'CLCFS' should be 'GLCFS'

src/main/webapp/js/models/GLCFSCollection.js
```diff
@@ -0,0 +1,160 @@
...
+						fetchSiteDataDeferred.reject(jqXHR);
+					}
+					else {
+						xmlResponse = xml;
+						fetchSiteDataDeferred.resolve(jqXHR);
+					}
+				},
+				error : function(jqXHR) {
+					log.debug('CLCFS GetFeature fetch failed');
```
mbucknell: Same typo

src/main/webapp/js/models/GLCFSCollection.js
```diff
@@ -0,0 +1,160 @@
...
+				});
+
+			return fetchDeferred.promise();
+		},
+
+		/*
+		 * @override
+		 * Need to pass the collection's timeBounds attribute to the getSelectedUrlParams method.
+		 */
+		getSelectedVariablesUrlParams : function() {
```
mbucknell: We no longer use this function and I failed to remove it from BaseDatasetCollection and PrecipitationCollection.

src/main/webapp/js/models/WorkflowStateModel.js
```diff
@@ -51,6 +52,11 @@ define([
 			}
 			else {
 				datasetCollections = _.object([
+				    [Config.GLCFS_ERIE_DATASET, new GLCFSCollection([],{lake:'Erie'})],
```
mbucknell: Should be using tabs for indenting.

src/main/webapp/js/views/MapView.js
```diff
@@ -21,9 +21,15 @@ define([
 	var siteIcons = _.mapObject(Config.DATASET_ICON, function(value) {
 		return L.icon(value);
 	});
+	
+	var getGLCFSTitle = function(model) {
+		return model.get('name');
```
mbucknell: This should probably be the same as the getPrecipTitle. The GLCFS site model has no "name" property.

src/test/js/test-main.js
```diff
@@ -16,6 +16,10 @@ require.config({
 			precipWFSGetFeatureUrl : 'http:dummyservice/wfs/?service=wfs&amp;version=2.0.0',
 			cidaThreddsPrecipData : 'dodsC/fakedata'
 		},
+		'models/GLCFSCollection' : {
+			GLCFSWFSGetFeatureUrl : 'http:dummyservice/wfs/?service=wfs&amp;version=2.0.0',
```
mbucknell: Probably want to mimic the app and assign different url's for each lake.

mbucknell:
Looks good! Still needs tests.

src/main/webapp/js/models/GLCFSCollection.js
instantsoup: Variables are going to be hardcoded/loaded from somewhere other than thredds since they never change.

instantsoup:
Fixes and updates, no tests yet.

src/main/webapp/js/models/GLCFSCollection.js
mbucknell: I didn't explain myself well. Not the GLCFS variables, but the data part of the dds url for each lake.

mbucknell:
We should add tests for the GLCFS site markers in MapView but I'm okay with doing that as part of the next two tickets.

👍

PR #77, opened 2016-06-09
Enddat 224

mbucknell:
Implements the ability for the user to define their area of interest by drawing a rubber band box rather than specifying a location and radius.

Started by creating the module AOIModel. This is where the latitude, longitude and radius or aoiBox properties will be saved. The AOIModel defines functions to return information about what kind of AOI definition is being used as well as the bounding box defined. The WorkflowStateModel now defines an aoi property by default that is initialized to an empty AOIModel.

Changed the PROJ_LOC_STEP to SPECIFY_AOI_STEP. The first part of this workflow step is for the user to choose picking a location or drawing a box.  If the user picks drawing a box, the new view AOIViewBox is rendered rather than LocationView and in MapView, drawControls are added to the map rather than the project location marker.

In order to have the entire AOI defined in a single view, I moved the radius from Choose Data to the LocationView. The LocationView now only needs to have the AOIModel passed to it rather than the entire WorkflowStateModel.

The URL for the Choose Data step when the AOI box is used, will be #aoiBbox/south,west,north,south/...

### DIFF
--- a/src/main/webapp/js/views/DataDiscoveryView.js
+++ b/src/main/webapp/js/views/DataDiscoveryView.js
@@ -18,6 +18,8 @@ define([
 		VariableSummaryView, ProcessDataView, hbTemplate) {
 	"use strict";
 
+	var DEFAULT_RADIUS = 2;
+
 	var NAVVIEW_SELECTOR = '.workflow-nav';
 	var LOCATION_SELECTOR = '.location-panel';
 	var CHOOSE_SELECTOR = '.choose-panel';
@@ -256,9 +258,15 @@ define([
 
 		selectAOIDefinition : function(ev) {
 			var kind = $(ev.currentTarget).val();
+			var aoiModel = this.model.get('aoi');
 			this.$('.workflow-start-container').removeClass('in');
 			switch(kind) {
 				case 'location':
+					aoiModel.set({
+						latitude : '',
+						longitude : '',
+						radius : DEFAULT_RADIUS
+					});
 					if (!this.locationView) {
 						this.locationView = new LocationView({
 							el : $utils.createDivInContainer(this.$(LOCATION_SELECTOR)),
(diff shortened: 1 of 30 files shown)


src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -76,8 +90,8 @@ define([
 			if (this.mapView) {
 				this.mapView.remove();
 			}
-			if (this.locationView) {
-				this.locationView.remove();
+			if (this.aoiView) {
```
mbucknell: this.aoiView can contain either an instance of LocationView or an instance AOIBoxView or be undefined. This is set when selectAOIDefinition is called (via a DOM event)

src/main/webapp/js/views/MapView.js
```diff
@@ -109,19 +126,27 @@ define([
...
 			});
-			_.each(this.controls, function(control) {
+			_.each(this.defaultControls, function(control) {
 				self.map.addControl(control);
 			}, this);
 			_.each(this.siteLayerGroups, function(layerGroup) {
 				self.map.addLayer(layerGroup);
 			});
 
+			// Set up the map event handlers for the draw control to update the aoi model.
```
mbucknell: The event handlers for the drawing controls update the aoi Model. Event handlers for changes to this model are used to update the layer representing the box.

src/main/webapp/js/views/NavView.js
```diff
@@ -76,23 +100,9 @@ define([
 			}
 		},
 
-		_getChooseDataUrl : function(model) {
```
mbucknell: Moved this function outside of the view.

src/main/webapp/js/views/NavView.js
```diff
@@ -10,11 +10,34 @@ define([
...
+		var state = model.attributes;
+		var aoi = model.attributes.aoi;
+
+		var startDate = (state.startDate) ? '/startdate/' + state.startDate.format(Config.DATE_FORMAT_URL) : '';
+		var endDate = (state.endDate) ? '/enddate/' + state.endDate.format(Config.DATE_FORMAT_URL) : '';
+		var datasets = (state.datasets) ? '/dataset/' + state.datasets.join('/') : '';
+
+		var aoiFragment = '';
+		if (aoi.usingProjectLocation()) {
+			var latitude = (aoi.attributes.latitude) ? aoi.attributes.latitude : '';
```
ayan-usgs: :+1: On using a ternary operator. I had to look this one up. :smile:

ayan-usgs:
Looks good to go.

PR #78, opened 2016-06-14
adding data view for GLCFS

instantsoup:
created for review and feedback

still needs tests

### DIFF
--- a/src/main/webapp/js/views/GLCFSDataView.js
+++ b/src/main/webapp/js/views/GLCFSDataView.js
@@ -1,0 +1,46 @@
+/* jslint browser: true */
+
+define([
+	'underscore',
+	'Config',
+	'views/BaseDataView',
+	'hbs!hb_templates/glcfsData'
+], function(_, Config, BaseDataView, hbTemplate) {
+
+	"use strict";
+
+	/*
+	 * @constructs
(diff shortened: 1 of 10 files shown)


src/main/webapp/js/views/GLCFSDataView.js
```diff
@@ -0,0 +1,46 @@
...
+		render : function() {
+			var getContextForVariable = function(variableModel) {
+				var result = _.clone(variableModel.attributes);
+				result.startDate = variableModel.attributes.startDate.format(Config.DATE_FORMAT);
+				result.endDate = variableModel.attributes.endDate.format(Config.DATE_FORMAT);
+				result.id = variableModel.cid;
+				return result;
+			};
+
+			this.context.name = this.model.get('name');
```
mbucknell: I'm not seeing a 'name' property in the GLCFS model. I would suggest listing the lat lon as the name or just leaving it off.

mbucknell:
There are dataset specific tests in MapViewSpec, "Tests for workflow model event handlers" that should also be added for consistency.

mbucknell:
There should also be dataset specific tests added to VariableSummary (not sure if that was on your list of remaining things)

PR #80, opened 2016-06-17
Enddat 164

ayan-usgs:


src/main/webapp/js/models/ShapefileUpload.js
```diff
@@ -0,0 +1,22 @@
+/* jslint browser: true */
```
mbucknell: This is not being used currently. I would take it out. I believe this is copied from GDP. GDP uses shapefiles differently then we do therefore I suspect how we save the shapefile feature will be different.

src/main/webapp/WEB-INF/web.xml
```diff
@@ -91,6 +91,63 @@
 	</servlet-mapping>
 	
 	<servlet>
+		<servlet-name>shapefile-upload-servlet</servlet-name>
```
mbucknell: It would be great to put the comments for the various parameters in. This is a complicated servlet to set up and it helps to have the comments with the parameters.

src/main/webapp/WEB-INF/jsp/head.jsp
```diff
@@ -30,6 +30,7 @@
 	String glcfsWFSGetFeatureUrlSuperior = props.getProperty("enddat.endpoint.glcfs.wfsgetfeature.superior");
 	String glosThreddsGLCFSData = props.getProperty("enddat.glosthredds.glcfsdata");
 	String acisStnMetaUrl = props.getProperty("enddat.endpoint.acis.stnmeta");
+	String gdpGeoServerUrl = props.getProperty("gdp.default.upload.workspace");
```
mbucknell: I don't see this used anywhere. Can we take it out?

src/main/webapp/js/views/DataDiscoveryView.js
```diff
@@ -310,6 +311,21 @@ define([
 						this.aoiView.expand();
 					}
 					break;
+				
+				case 'aoiShpFile':
```
mbucknell: I would take this out entirely since its not in the template. This code is straightforward and will probably not be called from DataDiscoveryView in the end.

src/main/webapp/js/views/ShapefileUploadView.js
```diff
@@ -0,0 +1,105 @@
...
+				type: 'POST',
+				dataType: 'xml',
+				send : function(e, data) {
+					data.url = data.url + '&qqfile=' + data.files[0].name;
+					log.info('Data URL: ' + data.url);
+					//$uploadIndicator.show();
+				},
+				done : function(e, data) {
+					//$uploadIndicator.hide();
+
```
mbucknell: Can you put a comment in here that the done and fail callbacks have not been tested? Better would be to put in log messages, but take out any code that affects the DOM or the models.

src/main/webapp/js/views/ShapefileUploadView.js
```diff
@@ -0,0 +1,105 @@
...
+					else {
+						var error = $resp.find('error').first().text();
+						var exception = $resp.find('exception').first().text();
+						self.alertView.show('alert-danger', 'File Upload error: ' + error + '. ' + exception);
+					}
+
+				},
+				fail : function(e, data) {
+					//$uploadIndicator.hide();
+					self.alertView.show('alert-danger', 'Upload failed');
```
mbucknell: Change this to a log statement.

mbucknell:
See the comments.  Otherwise it looks good to merge.

src/main/webapp/WEB-INF/web.xml
ayan-usgs: Adapted comments from gpd_web.

ayan-usgs:
Adjusted code with feedback.

PR #81, opened 2016-06-17
splash page start

mwernimont:
@mbucknell Theres still some testing on bigger and smaller screens but for the most part this is basically good to go and thought I should check it in. Lemme know if you see any issues,will do my best to fix em.


mbucknell:
I think its missing the change to custom.less to import the splash.less page?

mwernimont:
@mbucknell Like that?

mbucknell:
Yes. Hopefully none of the selectors in splash.less will interfer with the rest of the app. We'll find out soon enough!

PR #82, opened 2016-06-17
add shim for jquery-ui

ayan-usgs:


ayan-usgs:
Not sure if this is the solution to why build 39 failed... thoughts?

mbucknell:
This would not have fixed the build. The issue was that the dependencies for ShapeFileuploadView need to be added to src/test/js/karma.conf.js and src/test/js/test-main.js. Instead of doing that, removed the dependency on ShapeFileuploadView from DataDiscoveryView. Now this code is not pulled into the app or the tests. See https://github.com/USGS-CIDA/enddat_web/pull/84 for the changes that I made. I'm going to close this pull request.

mbucknell:
It's a good idea to do a clean build (mvn clean install) prior to opening a pull request. If there are any problems with the tests or forgetting to add dependencies to bower.json, this will catch it.

PR #91, opened 2016-06-27
GA for Dataset selection

mwernimont:
@mbucknell heres my code for the GA tracker of the selection for datasets. 

I first had it in the selectDataset function previously created, but was receiving multiple calls of the function which was reporting the selection text to google too many times causing fake data to appear.

So I created the gaSelectorTracker function, do you see any issues with that? It seems to not be messing with anything from what I've seen, but if you can see a cleaner solution I'm all ears.

### DIFF
--- a/src/main/webapp/js/views/ChooseView.js
+++ b/src/main/webapp/js/views/ChooseView.js
@@ -57,6 +57,7 @@ define([
 			this.updateDatasets();
 			this.updateStartDate();
 			this.updateEndDate();
+                        this.gaSelectTracker();
 
 			this.listenTo(this.model, 'change:datasets', this.updateDatasets);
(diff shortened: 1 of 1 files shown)


mbucknell:
I'm not sure why this would work any different than if you but the call to ga in the selectDatasets. But the code seems fine and it does isolate it from our tests, so I'm good with it.

One question I do have is is it a problem for us to be calling ga from our local development as well as our qa and dev tiers? Do these get filtered out somehow?

mwernimont:
Well the GA is looking for a URL to track, so it shouldn't be tracking it unless it matches the URL right? Or what do you mean?

mbucknell:
It's been a while since I've really looked at Google Analytics and how it works (it is a moving target!) and wasn't sure. I'm fine with this as solution.

PR #92, opened 2016-06-27
cleanup for GLCFS

instantsoup:


src/main/webapp/js/hb_templates/glcfsData.hbs
```diff
@@ -1,5 +1,6 @@
 <ul class="data-view-info">
-	<li><b>{{lake}}</b></li>
+	<li><b>Lat:&nbsp;</b>{{lat}}</li>
```
mbucknell: I'll let Jessica have the final say, but I would prefer Location: {{lat}}, {{lon}}

mbucknell:
👍

PR #94, opened 2016-06-29
Geoserver

mbucknell:
Finished implementation of ENDDAT-164. The cida-owsutils-geoserver version needed to be updated to the latest (1.0.6) in order to bring in the correct dependencies for the cida proxy.

When the shapefile is uploaded successfully, the code buts a property on the workflow state model containing the feature name to be used when displaying on the map. The MapView added an event handler for this property in order to draw (or remove the previous) uploaded feature.

### DIFF
--- a/src/main/webapp/WEB-INF/web.xml
+++ b/src/main/webapp/WEB-INF/web.xml
@@ -105,7 +105,7 @@
 		<init-param>
 			<!-- URL of the geoserver instances receiving the upload. -->
 			<param-name>geoserver.endpoint</param-name>
-			<param-value></param-value>
+			<param-value />
 		</init-param>
 		<init-param>
 		<!-- geoserver username -->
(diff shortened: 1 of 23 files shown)


src/main/webapp/WEB-INF/web.xml
```diff
@@ -105,7 +105,7 @@
 		<init-param>
 			<!-- URL of the geoserver instances receiving the upload. -->
 			<param-name>geoserver.endpoint</param-name>
-			<param-value>http://cida-eros-gdpdev.er.usgs.gov:8082/geoserver</param-value>
+			<param-value></param-value>
```
ayan-usgs: Would it be more stylistically consistent if this used a self-closing tag `<param-value />`?

ayan-usgs:
:+1:

Looks great! Must admit I would have struggled to put all the puzzle pieces together. :stuck_out_tongue:

PR #95, opened 2016-07-06
Enddat 229

ayan-usgs:
@mbucknell:

Could you give this a look and let me know what you think of the additions?

### DIFF
--- a/src/main/webapp/js/views/ProcessDataView.js
+++ b/src/main/webapp/js/views/ProcessDataView.js
@@ -28,8 +28,8 @@ define([
 	
 	var constructClassifier = function(param) {
 		var name = param.name;
-		var site_id = param.site_id;
-		var classifier = name + '--' + site_id;  // make a simple string to identify each site type and site number pair
+		var siteNo = param.siteNo;
+		var classifier = name + '--' + siteNo;  // make a simple string to identify each site type and site number pair
 		return classifier;
 	};
 	
--- a/src/main/webapp/js/views/ProcessDataView.js
+++ b/src/main/webapp/js/views/ProcessDataView.js
@@ -20,7 +20,7 @@ define([
 
 	var BASE_URL = module.config().baseUrl;
 	
-	var URL_LENGTH = 200;
+	var URL_LENGTH = 2000; // max url character length before it gets broken down into urls by site
 	
 	var constructClassifier = function(param) {
 		var name = param.name;
@@ -36,6 +36,7 @@ define([
 		// organize parameters by their sites
 		for (var i = 0; i < classifiers.length; i++) {
 			var paramClassifier = classifiers[i];
+			// make a list for parameters that fall into a specific dataset/site pair
 			var classifierParams = _.filter(params, function(param) {return constructClassifier(param) == paramClassifier});
 			masterParams.push(classifierParams);
 		}
@@ -211,7 +212,7 @@ define([
 			ev.preventDefault();
 			this.context.dataUrls = dataUrls;
 			var template = urlContainer;
-			$('.url-container').html(template({dataUrls : dataUrls}));
+			$('.url-container').html(template({dataUrls : dataUrls})); // render content in the url-container div
 			var $getDataBtn = this.$('.get-data-btn');
 			var $downloadBtn = this.$('.download-data-btn');
 			var $message = this.$('.warning-msg');
--- a/src/main/webapp/js/views/ProcessDataView.js
+++ b/src/main/webapp/js/views/ProcessDataView.js
@@ -30,11 +30,9 @@ define([
 	var organizeParams = function(params) {
 		// build array of identifying site type and number pairs
 		var classifiers = _.uniq(_.map(params, constructClassifier));
-		// make an array of arrays... each sub-array is site specific 
-		var masterParams = _.map(classifiers, function(classifier) {
-			var classifierParams = _.filter(params, function(param) {return constructClassifier(param) == classifier});
-			return classifierParams;
-		});
+		// make an object containing of parameters for each site
+		// e.g. {site1 : [parameters for site 1], site2 : [parameters for site2], ...}
+		var masterParams = _.groupBy(params, constructClassifier);
 		return masterParams;
 	};
 
@@ -64,9 +62,11 @@ define([
 		var siteUrls;
 		if (urlLength > maxUrlLength) {
 			var siteOrganizedParams = organizeParams(varParams);
-			siteUrls = _.map(siteOrganizedParams, function(siteParams) {
+			// take the site organizied parameters and create a url for each site,
+			// then return the values from the new object as an array
+			siteUrls = _.chain(siteOrganizedParams).mapObject(function(siteParams) {
 				return BASE_URL + 'service/execute?' + $.param(params.concat(siteParams));
-				});
+				}).values().value();
 		}
 		else {
 			siteUrls = [dataProcessingUrl];
@@ -105,7 +105,7 @@ define([
 		
 		initialize : function(options) {
 			BaseCollapsiblePanelView.prototype.initialize.apply(this, arguments);
-			this.maxUrlLength = options.maxUrlLength ? options.maxUrlLength : 2000; // max url character length before it gets broken down into urls by site
+			this.maxUrlLength = options.maxUrlLength ? options.maxUrlLength : 200; // max url character length before it gets broken down into urls by site
 		},
 
 		render : function() {
--- a/src/main/webapp/js/views/ProcessDataView.js
+++ b/src/main/webapp/js/views/ProcessDataView.js
@@ -82,7 +82,6 @@ define([
 		if (download) {
 			params.push({name : 'download', value: 'true'});
 		}
-		console.log(varParams);
 		var dataProcessingUrl = BASE_URL + 'service/execute?' + $.param(params.concat(varParams));
 		var urlLength = dataProcessingUrl.length;
 		if (urlLength > URL_LENGTH) {
--- a/src/main/webapp/js/views/ProcessDataView.js
+++ b/src/main/webapp/js/views/ProcessDataView.js
@@ -105,7 +105,7 @@ define([
 		
 		initialize : function(options) {
 			BaseCollapsiblePanelView.prototype.initialize.apply(this, arguments);
-			this.maxUrlLength = options.maxUrlLength ? options.maxUrlLength : 200; // max url character length before it gets broken down into urls by site
+			this.maxUrlLength = options.maxUrlLength ? options.maxUrlLength : 2000; // max url character length before it gets broken down into urls by site
 		},
 
 		render : function() {
--- a/src/main/webapp/js/views/ProcessDataView.js
+++ b/src/main/webapp/js/views/ProcessDataView.js
@@ -19,6 +19,8 @@ define([
 
 	var BASE_URL = module.config().baseUrl;
 	
+	var URL_LENGTH = 200;
+	
 	var isInArray = function(value, array) {
 		// determine if a value is in an array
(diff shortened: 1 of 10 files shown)


ayan-usgs:
Still need to add some tests before I put in a PR for an actual merge.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -18,6 +18,47 @@ define([
 	"use strict";
 
 	var BASE_URL = module.config().baseUrl;
+	
+	var isInArray = function(value, array) {
```
mbucknell: Consider using underscore's contains method - http://underscorejs.org/#contains

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -18,6 +18,47 @@ define([
...
+		// determine if a value is in an array
+		return array.indexOf(value) > -1;
+	};
+	
+	var constructClassifier = function(param) {
+		var name = param.name;
+		var value = param.value;
+		var siteNumber = value.split(":")[0];
```
mbucknell: Unfortunately this will not work if you want to uniquely identify a variables dataset/site. For instance for precip the value will be Precip=580:716:126517:precip3 and its 580:716 which uniquely identifies a site.

I think at the moment we can organize by dataset which should be uniquely identified by the 'name' property.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -18,6 +18,47 @@ define([
...
+		var classifier = name + '--' + siteNumber;  // make a simple string to identify each site type and site number pair
+		return classifier;
+	};
+	
+	var organizeParams = function(params) {
+		var classifiers = [];
+		// build list of identifying site type and number pairs
+		for (var i = 0; i < params.length; i++) {
```
mbucknell: You should look at the underscore library to simplify the array processing - http://underscorejs.org. For instance, you can use _.map and _.filter to get the list of classifiers (can even use _.chain).

Even fancier would be to use _.map to add the classifier to the object, use _.groupBy to group by the classifier rather than doing these two loops.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -18,6 +18,47 @@ define([
...
+				if (objectClassifier == paramClassifier) {
+					classifierParams.push(paramObject);
+				}
+			}
+			masterParams.push(classifierParams);
+		}
+		return masterParams;
+	};
 
 	var getUrl = function(workflowModel, download) {
```
mbucknell: Consider changing the name to getUrls - indicates that it could be more than one.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -40,10 +81,23 @@ define([
 		if (download) {
 			params.push({name : 'download', value: 'true'});
 		}
-
-		return BASE_URL + 'service/execute?' + $.param(params.concat(varParams));
+		var dataProcessingUrl = BASE_URL + 'service/execute?' + $.param(params.concat(varParams));
+		var urlLength = dataProcessingUrl.length;
+		if (urlLength > 200) {
```
mbucknell: You should make the length a module constant. Ultimately this should be 2000.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -40,10 +81,23 @@ define([
...
+			var siteOrganizedParams = organizeParams(varParams);
+			var siteUrls = [];
+			for (var i = 0; i < siteOrganizedParams.length; i++) {
```
mbucknell: Consider using _.map

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -176,11 +230,21 @@ define([
...
 
 		showUrl : function(ev) {
-			var url = getUrl(this.model);
-			var $link = this.$('.url-container a');
+			var dataUrls = getUrl(this.model);
 			ev.preventDefault();
-
-			$link.attr('href', url).html(url);
+			this.context.dataUrls = dataUrls;
+			BaseCollapsiblePanelView.prototype.render.apply(this, arguments);
```
mbucknell: The entire view should not be rendered when pushing the show button. You will lose all of your input settings. One option would be to create a Handlebars template (could probably hard code this one in this module) to render just the url display.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -176,11 +230,21 @@ define([
...
+			if (dataUrls.length > 1) {
```
mbucknell: There should be an else part to this that clears the warning message and reenables the Get Data and Download buttons. It's possible (probably not likely) that a user could decrease the url character count by eliminating some of the TIme Series Processing Options. Each one of those adds a new parameter to the url.

mbucknell:
If the url for a single dataset exceeds the limit, I think we should start dividing it into a url that handle a certain number of variables (say 10). To separate it by site, I think we would have to add another property to VariableParameter for siteId. I think the former would be easier to implement but I leave that up to you.

src/main/webapp/js/views/ProcessDataView.js
ayan-usgs: Would rendering the separate template require another view?

mbucknell: No ... you can use Handlebars at any point so you could define your template:

templateStr = '<p>{{myText}}'

Compile your template
template = Handlebars.compile(templateStr)

and then render it where you want

$('body').html(template({myText : 'Hello'})

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -12,14 +12,38 @@ define([
...
+	'hbs!hb_templates/processData',
+	'hbs!hb_templates/urlContainer'
 ], function($, _, moment, Handlebars, datetimepicker, stickit, module, Config, $utils, BaseCollapsiblePanelView,
-	VariableTsOptionView, hbTemplate) {
+	VariableTsOptionView, hbTemplate, urlContainer) {
 	"use strict";
 
 	var BASE_URL = module.config().baseUrl;
+	
+	var URL_LENGTH = 2000; // max url character length before it gets broken down into urls by site
```
mbucknell: It might make testing easier if the URL_LENGTH was not so long. Consider passing it as an option to the view. It can default to 2000, but this way when you create the view to test, you can have a smaller limit.

ayan-usgs:
@mbucknell,

I've implemented the changes you suggested and added some tests. Could you give this a look before I merge?

Thanks!

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -12,14 +12,36 @@ define([
...
+		var classifier = name + '--' + siteNo;  // make a simple string to identify each dataset type and site number pair
+		return classifier;
+	};
+	
+	var organizeParams = function(params) {
+		// build list of identifying site type and number pairs
+		var classifiers = _.uniq(_.map(params, constructClassifier));
+		var masterParams = [];
+		// organize parameters by their sites
+		for (var i = 0; i < classifiers.length; i++) {
```
mbucknell: Consider at least using _.map to replace the for loop. If you have some time to kill, you might want to investigate using _.groupBy.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -40,10 +62,18 @@ define([
 		if (download) {
 			params.push({name : 'download', value: 'true'});
 		}
-
-		return BASE_URL + 'service/execute?' + $.param(params.concat(varParams));
+		var dataProcessingUrl = BASE_URL + 'service/execute?' + $.param(params.concat(varParams));
+		var urlLength = dataProcessingUrl.length;
+		if (urlLength > maxUrlLength) {
+			var siteOrganizedParams = organizeParams(varParams);
+			var siteUrls = _.map(siteOrganizedParams, function(siteParams) {return BASE_URL + 'service/execute?' + $.param(params.concat(siteParams))});
```
mbucknell: Not a big fan of putting the function body on the same line as the anonymous function declaration, particularly when the line then exceeds the line limit in git

mbucknell: Also you should declare var siteUrls outside of the if block. I try to declare all of my variables in a function prior to any executable code.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -176,21 +211,39 @@ define([
...
 
 		showUrl : function(ev) {
-			var url = getUrl(this.model);
-			var $link = this.$('.url-container a');
+			var dataUrls = getUrls(this.model, this.maxUrlLength);
 			ev.preventDefault();
-
-			$link.attr('href', url).html(url);
+			this.context.dataUrls = dataUrls;
+			var template = urlContainer;
```
mbucknell: This variable seems unnecessary. Maybe change the name of urlContainer to urlContainerTemplate

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -176,21 +211,39 @@ define([
...
+				// enable "Get data" and "Download" buttons
+				$getDataBtn.prop("disabled", false);
+				$downloadBtn.prop("disabled", false);
+			}
 		},
 
 		getData : function(ev) {
 			ev.preventDefault();
-			window.open(getUrl(this.model), '_blank');
+			window.open(getUrls(this.model, this.maxUrlLength), '_blank');
```
mbucknell: Does this work? getUrls now returns an array rather than a string. I would think you would need to pass getUrls(...)[0].

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -176,21 +211,39 @@ define([
...
 		},
 
 		downloadData : function(ev) {
 			ev.preventDefault();
-			window.location.assign(getUrl(this.model, true));
+			window.location.assign(getUrls(this.model, this.maxUrlLength, true));
```
mbucknell: Same comment as above

src/test/js/views/ProcessDataViewSpec.js
```diff
@@ -64,11 +64,12 @@ define([
 					}
 				});
 
-				spyOn(testModel, 'getSelectedVariables').and.returnValue(variableCollection.models);
+				spy = spyOn(testModel, 'getSelectedVariables').and.returnValue(variableCollection.models);
```
mbucknell: You could probably move this spy to the tests for "DOM events for processing buttons with long URL. Make sure you declare it as var spy, otherwise you are putting in a global. I forget to put "use strict"; after the function(..) at the top of the spec, but that should catch that you are not declaring a variable. Also JSHint in your IDE might help with things like that.

mbucknell: never mind ... I'm tired! Maybe give this a more descriptive name than spy though.

src/test/js/views/ProcessDataViewSpec.js
```diff
@@ -237,7 +238,77 @@ define([
...
+							end : moment('2006-06-30', Config.DATE_FORMAT)
+						}
+					});				
+				});
+				
+				it('Expects a message be shown for more than one site url', function() {
+					$testDiv.find('.show-url-btn').trigger('click');
+					var expectedMsg = 'The URL for data processing exceeds the character limit. A single URL has been provided for each selected station.'
+					var message = $testDiv.find('p.warning-msg').html();
+				    expect(message).toEqual(expectedMsg);
```
mbucknell: Looks like spaces rather than tab issue here with indentation.

mbucknell:
Nice tests! Just a few things, mostly stylistic. 👍

PR #98, opened 2016-07-11
Enddat 229

ayan-usgs:
Not ready to merge yet, but I'd appreciate some help on a test.


src/test/js/views/ProcessDataViewSpec.js
```diff
@@ -81,6 +83,69 @@ define([
...
+				var isDisabled = $testDiv.find('.download-data-btn').is(':disabled');
+				expect(isDisabled).toBe(true);
+			});
+			
+			it('Expects the get data button is disabled.', function() {
+				var isDisabled = $testDiv.find('.get-data-btn').is(':disabled');
+				expect(isDisabled).toBe(true);
+			});
+			
+			it('Expects that buttons are enabled upon unchecking TS options after the view has rendered.', function() {
```
ayan-usgs: On this particular test, I'm trying to alter the model and then verify that with fewer variables selected the buttons are enabled. I put in some log statements, but they the changes don't seem be caught.

@mbucknell: Do you have any suggestions?

src/test/js/views/ProcessDataViewSpec.js
```diff
@@ -81,6 +83,69 @@ define([
...
+				// test that unselecting time-series variables will enable the buttons
+				// unselect the last two variables...
+				_.each(variableCollectionLong.models.slice(1), function(varModel) {
```
mbucknell: I'm not understanding what you are doing here. But I think you are trying to update the result of getSelectedVarSpy to include fewer models. I think the following should work:

getSelectedVarSpy.and.returnValue(variableCollection.models.slice(1). This call to slice should return an array without the first element of variableCollection. Note that this does not update the variableCollection variable. It remains the same.

ayan-usgs: I'm trying to test the behavior where a user comes into the processing page with a lot of variables, and hence a long URL and the buttons are disabled; then the user deselects them on the processing page, resulting in a shorter URL and the buttons are enabled. 

I was thinking the way to simulate that would be to alter a subset of the models in the `variableCollection` to trigger the listener on `change:timeSeriesOptions`; it also seemed that for this to work, the returnValue on the spy would also have to be changed.

mbucknell: Yes, one way to do this is to aleter what is returned from the spy. However, in this code you are not altering it, you continue to return variableCollectionLong which has not been modified. Maybe try something like this:

getSelectedVarSpy.and.returnValue(variableCollectionLong.models.slice(1)

variabelCollectionLogn.models.slice(1) will return a new array containing all the elements in variableCollectionLong.models except the first element. You shouldn't need the _.each at all.

ayan-usgs: Hmm... in light of our discussion, I'm considering removing the test. It doesn't seem like it would be redundant with the two specs starting at https://github.com/USGS-CIDA/enddat_web/pull/98/files#diff-2b2844e281ad7097d3b7efc78d67b181R183.

src/test/js/views/ProcessDataViewSpec.js
```diff
@@ -4,14 +4,16 @@
...
 	'moment',
 	'Config',
 	'bootstrap-datetimepicker',
 	'utils/VariableParameter',
 	'models/WorkflowStateModel',
 	'models/BaseVariableCollection',
-	'views/BaseView'
-], function(Squire, $, moment, Config, datetimepicker, VariableParameter, WorkflowStateModel, BaseVariableCollection, BaseView) {
+	'views/BaseView',
+	'views/VariableTsOptionView'
```
mbucknell: This is not necessary. This view is being injected into ProcessDataView with Squire (see line 36).

src/test/js/views/ProcessDataViewSpec.js
mbucknell: Since the TS options are actually part of a separate view (the VariableTsOptionView) which we are stubbing out with Squire, this test is unnecessary. This logic should be tested by updating the variables themselves (selecting or unselecting) and then verifying that the event handler correctly updates the disabled state of the button.

mbucknell: What you may want to add is a test where the variable length is already exceeded when render is called. You will need a new describe block since you'll have to set the value of testModel appropriately before testView.render is called.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -146,9 +150,15 @@ define([
...
 				maxDate : selectedVarsDateRange.end
 			});
-
+			// do the initial check for URL length when rendering after the user selects data
+			this.urlLengthBtnControl();
 			this.listenTo(this.model, 'change:outputDateRange', this.updateOutputDateRangeInputs);
-
+			
+			var timeSeriesOptionsModels = this.model.getSelectedVariables();
+			_.each(timeSeriesOptionsModels, function(timeSeriesOptionsModel) {
```
mbucknell: This could be added at the same time that the VariableTsOptionView is initialized. That way you don't have to iterate over the selected variable models twice.

src/test/js/views/ProcessDataViewSpec.js
ayan-usgs: I'm confused. I think this suite already start off with a set of variables that would exceeds the max URL length of 215. I'm not sure I understand why a new describe block is needed. Can you elaborate?

mbucknell: If that is true then you don't, although then you might want to add a test to make sure that they are enabled if the url length is under 215.

ayan-usgs:
@mbucknell:

I finally got that test working. Could you give this PR a final look before I merge?

Thanks!

mbucknell:
Wonderful! Good to go! 👍

PR #108, opened 2016-07-15
Enddat 238

mbucknell:


src/main/webapp/js/controller/AppRouter.js
```diff
@@ -72,7 +73,24 @@ define([
 		},
 
 		chooseDataState : function(workflowStep, state) {
+			var datasetsInUrl = state.datasets ? state.datasets.split('/') : [];
+			var glcfsLake = '';
+			var datasets = [];
```
mbucknell: Needed to be able to determine if the datasets in the url fragment included 'GLCFS_lakename', extract the lake name and set it in the datasetCollections

src/main/webapp/js/views/ChooseView.js
```diff
@@ -24,8 +26,9 @@ define([
 		panelBodyId : 'choose-data-panel-body',
 
 		additionalEvents : {
-			'select2:select #datasets-select' : 'selectDataset',
-			'select2:unselect #datasets-select' : 'resetDataset'
+			'select2:selecting #datasets-select' : 'selectDataset',
```
mbucknell: Had to use selecting rather than select because I can catch the select2 event before it updates the selections displayed.

src/test/js/views/ProcessDataViewSpec.js
```diff
@@ -322,104 +325,118 @@ define([
 				expect(testModel.get('outputTimeGapInterval')).toEqual('12');
 			});
 
-		describe('DOM events for processing buttons with a long URL', function() {
```
mbucknell: A lot of the remaining changes seem to be whitespace issues.

src/test/js/views/ProcessDataViewSpec.js
```diff
@@ -211,35 +215,34 @@ define([
...
+				datasetCollections = testModel.get('datasetCollections');
+				datasetCollections[Config.PRECIP_DATASET].reset([{variables : variableCollectionLong}]);
 				spyOn(testModel, 'getSelectedVarsDateRange').and.returnValue({
 					start : moment('2000-01-04', Config.DATE_FORMAT),
 					end : moment('2005-06-01', Config.DATE_FORMAT)
 				});
-				testModel.initializeDatasetCollections();
-				datasetCollections = testModel.get('datasetCollections');
-				datasetCollections[Config.PRECIP_DATASET].reset([{variables : variableCollectionLong}]);
-				testModel.getSelectedVariables.and.callThrough();
```
ayan-usgs: Huh... the tests still work with this removed (I just tried it). Confused as to why it worked, but that's pretty cool.

src/main/webapp/js/views/ChooseView.js
ayan-usgs: I'm wondering if this should be a comment in the code.

mbucknell: That would be helpful! I'll add that.

src/test/js/views/ChooseViewSpec.js
```diff
@@ -1,14 +1,16 @@
 /* jslint browser: true */
-/* global jasmine, expect, sinon */
+/* global jasmine, expect, spyOn */
 
 define([
 	'squire',
 	'jquery',
 	'select2',
+	'bootstrap',
```
ayan-usgs: Is this necessary? I don't see it being used in this module.

ayan-usgs: Nevermind. Found it.

ayan-usgs:
:+1:

src/test/js/views/ProcessDataViewSpec.js
mbucknell: I think for two reasons, I don't thinks we are spying on getSelectedVariables at this point and you are telling it to callThrough which is the normal behavior.

ayan-usgs:
Sorry... those conflicts are coming from my merge earlier this morning... :frowning:

src/test/js/views/ProcessDataViewSpec.js
ayan-usgs: Interesting. I was thinking that the spyOn at https://github.com/USGS-CIDA/enddat_web/pull/108/files#diff-2b2844e281ad7097d3b7efc78d67b181L89 was global.

mbucknell: You are right ...  missed that. I think it's working because the test only tests to see is the button whether it is enabled at the end of the test. I think it's enabled at the beginning of the test and remains enabled. I'll modify that test to make sure its testing what we want.

ayan-usgs: I think it should be fine if `callThrough()` is added back. Thoughts?

mbucknell: Yes, but it may be better to put in in the specific test so that we can test changing it to disable and then back to enabled.

PR #110, opened 2016-07-18
get JQUERY path to work

mwernimont:
This should make the tabs work for the intro page. Right @jiwalker-usgs ?

It at least made my local version work again

or @mbucknell ?


mbucknell:
It should. Merge away

PR #112, opened 2016-07-19
Enddat 232

mbucknell:
The file app.build.js is passed to the require optimizer. The optimizer is run as part of the build process using the requirejs-maven-plugin.  The application will used the optimized javascript if the JNDI variable enddat.development is set to false.

Also we are now compressing the compiled less. When developers want to see uncompressed css, either compile locally or set the compress option in the maven plugin to false.


app.build.js
```diff
@@ -0,0 +1,58 @@
...
+		}
+	],
+
+	name : 'init',
+
+	deps : ['../bower_components/requirejs/require'],
+
+	insertRequire: ["init"],
+
+	dir: "src/main/webapp/assets-build"
```
mbucknell: Not sure this is the best place to put the compressed javascript. Other suggestions are welcome.

PR #117, opened 2016-07-21
adding more variables to the variableDatasetMapping

dblodgett-usgs:
I used the var major code for snow water equivalent... not sure that will work?
See table 3: http://www.rcc-acis.org/docs_webservices.html for what I mean by var major.


mbucknell:
The rule is if it has code in the table use that otherwise specify the var major value as the code property's value. So I think this is good.

PR #124, opened 2016-07-26
Fix tiles

mbucknell:


mbucknell:
Also updated the version of the frontend-maven-plugin and the version of node to the latest LTS release.

PR #127, opened 2016-07-27
Enddat 244

ayan-usgs:
@mbucknell: Any testing suggestions would be most welcome. I'm a bit flummoxed about how to test for a successful download. :pensive:


mbucknell:
I'll review probably tomorrow morning. But I don't think you can really test for a successful download. But you can spy on Blob and saveAs to see that what you are passing to them is what you expected. We did this in NWC: https://github.com/USGS-CIDA/nwc/blob/master/nwc/src/test/javascript/specs/WaterbudgetPlotView_spec.js#L120

mbucknell:
Started looking at this. I'm not crazy about putting more site information in variableParameter in each variable. I think if you add a getSitesWithSelectedVariables to WorkflowStateModel and use this when someone requests meta data information, I think you can avoid putting more site information in the variable model.

ayan-usgs:
Thanks for the suggestion. I've give that a go.

src/main/webapp/js/models/ACISCollection.js
```diff
@@ -59,8 +59,14 @@ define([
...
 								siteNo : sid,
+								siteName : site.name,
 								value : sid + ':' +  result.code,
-								colName : result.description + ':' + sid
+								colName : result.description + ':' + sid,
+								latitude : site.ll[1],
+								longitude : site.ll[0],
+								elevation : site.elev,
+								elevationUnit : 'ft',
+								variableUnit : result.unit
```
mbucknell: You don't need to put the variableUnit in variableParameter. It's already in the variable's model's property.

src/main/webapp/js/models/NWISCollection.js
```diff
@@ -95,8 +101,14 @@ define([
 									variableParameter : new VariableParameter({
 										name : 'NWIS',
 										siteNo: variable.site_no,
+										siteName : variable.station_nm,
 										value : variable.site_no + ':' + variable.parm_cd + ':' +  statCd,
-										colName : name + ':' + variable.site_no
+										colName : name + ':' + variable.site_no,
+										latitude : variable.dec_lat_va,
+										longitude : variable.dec_long_va,
+										elevation : null,
```
mbucknell: I think NWIS does have elevation information. It's the alt_va column and the alt_datum_cd column is the "units". You can find information about the NWIS site service here: http://waterservices.usgs.gov/rest/Site-Service.html

src/main/webapp/js/models/NWISCollection.js
```diff
@@ -147,17 +159,19 @@ define([
 			var deferred = $.Deferred();
 			$.ajax({
 				type : "GET",
-				url : 'pmcodes/?radio_pm_search=param_group&pm_group=Physical&format=rdb&show=parameter_nm',
+				url : 'pmcodes/?radio_pm_search=param_group&pm_group=Physical&format=rdb&show=parameter_nm&show=parameter_units', // output is a bit cleaner than show=parameter_nm,parameter_units
 				dataType: 'text',
 				success: function(data) {
 					var parsedParams = rdbUtils.parseRDB(data);
 					self.parameterCodes = _.object(_.pluck(parsedParams, 'parameter_cd'), _.pluck(parsedParams, 'parameter_nm'));
+					self.parameterUnits = _.object(_.pluck(parsedParams, 'parameter_cd'), _.pluck(parsedParams, 'parameter_units'));
```
mbucknell: Consider putting the parsed parameterUnits in the parameterCodes object.

mbucknell:
I'll wait on the changes you are making before doing a full review.

src/main/webapp/js/models/NWISCollection.js
ayan-usgs: > I think NWIS does have elevation information. It's the alt_va column and the alt_datum_cd column is the "units". You can find information about the NWIS site service here: http://waterservices.usgs.gov/rest/Site-Service.html

Yeah, `alt_datum_cd` isn't the units per se, but one can figure it out with by doing some math. I'll leave that math to the end user.

ayan-usgs: I take that back. I don't think one can figure out units from the alt_datum_cd. The documentation implies the the unit is `ft` as measured using the specified datum.

ayan-usgs:
@mbucknell:

This is PR is ready for another look.

src/main/webapp/js/models/BaseDatasetCollection.js
```diff
@@ -39,6 +39,7 @@ define([
 		getSelectedVariables : function() {
 			return this.chain()
 				.map(function(datasetModel) {
+					var metaData = datasetModel.get('variables').getSelectedVariableMetadata();
```
mbucknell: This variable is not used.

src/main/webapp/js/models/BaseVariableCollection.js
```diff
@@ -69,6 +69,13 @@ define([
 
 			return result;
 		},
+		
+		getSelectedVariableMetadata : function() {
```
mbucknell: I think this is no longer used and can be removed.

src/main/webapp/js/models/GLCFSCollection.js
```diff
@@ -101,6 +102,11 @@ define([
 				result.push({
 					lon : $utils.xmlFind($this, 'sb', 'X1').text(),
 					lat : $utils.xmlFind($this, 'sb', 'X2').text(),
+					elevation : null,
+					elevationUnit : null,
+					siteNo : y + ':' + x,
+					name : y + ':' + x,
+					datasetName : datasetName,
```
mbucknell: I think it would be better to use GLCFS rather than GRID to identify the dataset. It's what we use in the application too.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -259,12 +281,71 @@ define([
...
+				var siteElevationUnit = siteModel.get('elevationUnit');
+				// get variable metadata
+				var varMetadata = _.map(siteVariables, function(siteVariable) {
+					var variableName = siteVariable.get('name') ? variableName : siteVariable.get('description');
+					var variableUnit = siteVariable.get('variableUnit');
+					var variableDataStr = 'variableName:' + variableName + ';variableUnit:' + variableUnit;
+					return variableDataStr;
+				})
+				.join('||');
+				var headers = ['dataset',
```
mbucknell: You can declare this constant outside of the _.map function. Probably should capitalize it.

src/main/webapp/js/models/BaseDatasetCollection.js
```diff
@@ -39,12 +39,34 @@ define([
...
 				})
 				.flatten()
 				.value();
 		},
+		/*
+		 * @ returns {Array of Backbone.Models} - each model represents a site with one or more selected variables
+		 */
+		
+		getSitesWithSelectedVariables : function() {
+			var selectedSites = _.map(this.models, function(datasetModel) {
```
mbucknell: I think if you use this.get('variables').hasSelectedVariables() and _.filter, youl'll get what you want.

ayan-usgs:
@mbucknell:

Ready for another look.

src/main/webapp/js/models/ACISCollection.js
```diff
@@ -57,7 +58,7 @@ define([
 							result.startDate = moment(dateRange[0], Config.DATE_FORMAT);
 							result.endDate = moment(dateRange[1], Config.DATE_FORMAT);
 							result.variableParameter = new VariableParameter({
-								name : 'ACIS',
+								name : datasetName,
 								siteNo : sid,
```
mbucknell: I believe we no longer need siteNo in the variableParameter, so these should be removed here and in the other dataset collections.

src/main/webapp/js/models/BaseDatasetCollection.js
```diff
@@ -44,6 +44,17 @@ define([
 				.flatten()
 				.value();
 		},
+		/*
+		 * @ returns {Array of Backbone.Models} - each model represents a site with one or more selected variables
+		 */
+		
+		getSitesWithSelectedVariables: function() {
+			var selectedSites = this.chain().filter(function(datasetModel) {
```
mbucknell: Does this chain need to be there? I only see the filter function.

src/main/webapp/js/models/GLCFSCollection.js
```diff
@@ -90,7 +91,7 @@ define([
 					var sigma = -1; // not using any 3D datasets currently
 					var vectorOpts = ':::0'; // not doing any vector options currently
 					siteVar.variableParameter = new VariableParameter({
-						name : 'GRID',
+						name : datasetName,
```
mbucknell: This name has to be "GRID" though because that is what the services expect.

src/main/webapp/js/models/WorkflowStateModel.js
```diff
@@ -79,7 +79,19 @@ define([
 				.flatten()
 				.value();
 		},
-
+		
+		/*
+		 * @returns {Array of arrays} - contains arrays of selected site for each dataset
```
mbucknell: Consider flattening this array of arrays here to just return an array of site models. From the returned value you can't really determine which dataset collection the sites came from.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -238,7 +275,8 @@ define([
 		},
 
 		showUrl : function(ev) {
-			var dataUrls = getUrls(this.model, this.maxUrlLength);
+			var dataUrls = _.values(getUrls(this.model, this.maxUrlLength));
```
mbucknell: Not sure what _.values is doing. Doesn't getUrls already return an array of strings?

ayan-usgs: It's not doing anything. It's a hold over from when `getUrls` briefly returned an object. I must have missed it while cleaning-up.

src/main/webapp/js/views/ProcessDataView.js
```diff
@@ -265,6 +303,55 @@ define([
 		downloadData : function(ev) {
 			ev.preventDefault();
 			window.location.assign(getUrls(this.model, this.maxUrlLength, true)[0]); // grab first item in the array
+		},
+		
+		provideMetadata : function(ev) {
+			var workflowModel = this.model;
+			var sitesWithSelectedVariablesByDataset = workflowModel.getSitesWithSelectedVariables();
+			var metaDataByDataset = _.map(sitesWithSelectedVariablesByDataset, function(datasetSelectedSites) {
+				var siteMetadata = _.map(datasetSelectedSites, function(datasetSelectedSite) {
```
mbucknell: You might consider declaring the map function here and then one creating the metadata has named function variables. It can make it easier to see what is happening.

mbucknell:
Just a few things. Otherwise looks good to go! 👍

src/main/webapp/js/views/ProcessDataView.js
ayan-usgs: Flattening out the return of `getSitesWithSelectedVariables` in the `WorkflowStateModel` simplified this a bit. I'm not sure that created a named function adds anymore clarity.

PR #128, opened 2016-07-29
Enddat 213

ayan-usgs:


mbucknell:
You should also add  a servletMapping in main/webapp/WEB-INF/web.xml so that we don't need to reference the page as UserGuide.jsp (probably userguide) . Also as part of this ticket, you should hook up the button in Intro.jsp that goes to the user guide (use userguide rather than UserGuide.jsp) and also hook up the Help button in the NavView in the data discovery page.

ayan-usgs:
> You should also add a servletMapping in main/webapp/WEB-INF/web.xml so that we don't need to reference the page as UserGuide.jsp (probably userguide) . Also as part of this ticket, you should hook up the button in Intro.jsp that goes to the user guide (use userguide rather than UserGuide.jsp) and also hook up the Help button in the NavView in the data discovery page.

Okay.

ayan-usgs:
@mbucknell:

The requested changes have been made.

PR #129, opened 2016-08-01
Refactor sites layer

mbucknell:
This is the first part of ENDDAT-248. The goal of this refactor was to put the site layer creation and management into its own view and to separate the site layer management into a module for the BySite workflow and another for the ByVariableType workflow.

Still need to add the new functionality for the ByVariableType workflow.

### DIFF
--- a/src/main/webapp/js/views/BySiteLayerView.js
+++ b/src/main/webapp/js/views/BySiteLayerView.js
@@ -43,6 +43,10 @@ define([
 			return this;
 		},
 
+		/*
+		 * Model event handlers
+		 */
+		
 		updateSiteMarkers : function() {
 			var self = this;
 			var dateFilter = this.model.has('dataDateFilter') ? this.model.get('dataDateFilter') : undefined;
--- a/src/main/webapp/js/views/BySiteLayerView.js
+++ b/src/main/webapp/js/views/BySiteLayerView.js
@@ -1,0 +1,80 @@
+/* jslint browser: true */
+
+define([
+	'underscore',
+	'leaflet',
+	'backbone'
+], function(_, L, Backbone) {
+	"use strict";
+	/*
+	 * This view will be rendered at initialization. A call to render will be a no op.
+	 * @constructs
+	 * @param {Object} options
+	 *		@prop {Leaflet.Map} map
+	 *		@prop {WorkflowStateModel} model
+	 *		@prop {Jquery element} el - This is the parent div of the div which contains the map
+	 *		@prop {String} datasetKind - The kind of collection
+	 *		@prop {L.Icon} siteIcon - The icon to be used for the collection's sites
+	 *		@prop {Function} getTitle - To be used when hovering over a site
+	 *			@param {Backbone.Model - represents a model in collection} model
+	 *			@returns String
+	 */
+	var view = Backbone.View.extend({
+
+		initialize : function(options) {
+			Backbone.View.prototype.initialize.apply(this, arguments);
+			this.collection = this.model.get('datasetCollections')[options.datasetKind];
+			this.datasetKind = options.datasetKind;
+			this.map = options.map;
+			this.siteIcon = options.siteIcon;
+			this.getTitle = options.getTitle;
+
+			this.siteLayerGroup = L.layerGroup();
+			this.map.addLayer(this.siteLayerGroup);
+
+			this.listenTo(this.model, 'change:dataDateFilter', this.updateSiteMarkers);
+			this.listenTo(this.collection, 'reset', this.updateSiteMarkers);
+			this.updateSiteMarkers();
+		},
+
+		remove : function() {
+			this.siteLayerGroup.clearLayers();
+			this.map.removeLayer(this.siteLayerGroup);
+			return this;
+		},
+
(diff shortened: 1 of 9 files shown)


src/main/webapp/js/views/BySiteLayerView.js
```diff
@@ -0,0 +1,84 @@
...
+			_.each(filteredSiteModels, function(siteModel) {
+				var latLng = L.latLng(siteModel.attributes.lat, siteModel.attributes.lon);
+				var marker = L.marker(latLng, {
+					icon : self.siteIcon,
+					title : self.getTitle(siteModel)
+				});
+
+				self.siteLayerGroup.addLayer(marker);
+				marker.on('click', function(ev) {
+					self.model.set('selectedSite', {
```
mbucknell: Setting selectedSite on the workflow model when the user clicks on a site. The SitesLayerView watches for changes on this property in order to display the appropriate DataView and to adjust the width of the map.

src/main/webapp/js/views/SitesLayerView.js
```diff
@@ -0,0 +1,194 @@
...
+
+			this.model.unset('selectedSite');
+			return this;
+		},
+
+		createSiteLayers : function() {
+			switch(this.model.get('step')) {
+				case Config.CHOOSE_DATA_BY_SITE_FILTERS_STEP:
+				case Config.CHOOSE_DATA_BY_SITE_VARIABLES_STEP:
+					this.listenTo(this.model, 'change:selectedSite', this.updateSelectedSite);
```
mbucknell: Setting up the event handler when workflow is By_Site to handle when the selectedSite property is updated.

src/main/webapp/js/views/ByVariableTypeLayerView.js
```diff
@@ -0,0 +1,69 @@
...
+	 *		@prop {Function} getTitle - To be used when hovering over a site
+	 *			@param {Backbone.Model - represents a model in collection} model
+	 *			@returns String
+	 *
+	 */
+	var view = Backbone.View.extend({
+		initialize : function(options) {
+			Backbone.View.prototype.initialize.apply(this, arguments);
+			this.collection = this.model.get('datasetCollections')[options.datasetKind];
+			this.datasetKind = options.datasetKind;
```
ayan-usgs: Doesn't totally affect functionality, but would this be more logical:

```
this.datasetKind = options.datasetKind;
this.collection = this.model.get('datasetCollections)[this.datasetKind];
```
It seems redundant to use `options.datasetKind` and then to define a variable using it in the next line.

mbucknell: I'm not sure there is a best practice for this. The "this.datasetKind" is used in another view method.

src/main/webapp/js/views/MapView.js
```diff
@@ -125,28 +76,27 @@ define([
...
 			}
-			this.map = L.map(this.mapDivId, {
+			this.map = L.map(MAP_DIV_ID, {
 				center: [41.0, -100.0],
 				zoom : 4,
 				layers : [this.baseLayers['World Street']]
 			});
 			_.each(this.defaultControls, function(control) {
 				self.map.addControl(control);
 			}, this);
```
ayan-usgs: Is passing `this` here really necessary? I would expect that `self` would maintain the reference to the original `this`.

src/test/js/views/MapViewSpec.js
```diff
@@ -260,46 +235,18 @@ define([
 				expect(L.map.calls.count()).toBe(2);
 			});
 
-			it('Expect that the site location marker is added to the map if there are sites in the site model', function() {
+			it('Expects that the sitesLayerView is created if the workflow step if CHOOSE_DATA_BY_SITE_FILTERS_STEP', function() {
```
ayan-usgs: Should this be `workflow step is CHOOSE_DATA_BY_SITE_FILTERS_STEP` ?

src/test/js/views/MapViewSpec.js
```diff
@@ -260,46 +235,18 @@ define([
...
 			});
 
-			it('Expects that the GLCFS layer group contains markers for each grid', function() {
-				testModel.set('step', Config.CHOOSE_DATA_BY_SITE_FILTERS_STEP);
-				testGLCFSCollection.reset([
-					{lon : '-100', lat : '43.0', variables : new BaseVariableCollection([{x : '1', y: '2'}])},
-					{lon : '-100', lat : '44.0', variables : new BaseVariableCollection([{x : '1', y: '3'}])}
-				]);
-				spyOn(testView.siteLayerGroups[Config.GLCFS_DATASET], 'addLayer').and.callThrough();
+			it('Expects that the sitesLayerView is created if the workflow step if CHOOSE_DATA_BY_VARIABLES_STEP', function() {
```
ayan-usgs: See above.

src/main/webapp/js/views/MapView.js
mbucknell: Nope, it's not necessary. I'll fix.

ayan-usgs:
:+1: 

Good for a merge at your discretion.

PR #130, opened 2016-08-02
Enddat 248

mbucknell:
Created a leaftlet select control and used this to implement filtering on the selected variable kind. The state of this filter is saved in the selectedVarKind in the workflow state model. The ByVariableTypeLayerView listens for changes of this property and updates the displayed sites for just this variable kind. Added a popup for each displayed site that contains a checkbox which can be used to select/unselect the variable.


src/main/webapp/js/views/SitesLayerView.js
```diff
@@ -185,6 +212,57 @@ define([
 				$dataViewDiv.removeClass(DATA_VIEW_WIDTH_CLASS);
 				this.map.invalidateSize();
 			}
+		},
+
+		updateVariableKindControl : function() {
```
ayan-usgs: I think some comments here might be helpful. Took me awhile to think through the different if/else statements.

ayan-usgs:
Good to go. :+1:

PR #142, opened 2016-12-14
Enddat 251

ayan-usgs:


src/main/webapp/js/views/MapView.js
```diff
@@ -134,6 +157,32 @@ define([
 				this.sitesLayerView = undefined;
 			}
 		},
+		
+		addGeoJsonLayer : function(dataPath, markerIcon) {
```
mbucknell: I would recommend calling this createGeoJsonLayer. The function does not add anything to the map which is what I assumed when I read the code. Also consider adding document comments for this function.

src/main/webapp/js/views/MapView.js
```diff
@@ -134,6 +157,32 @@ define([
...
+				onEachFeature: function (feature, layer) {
+					var featureName = feature.properties.name;
+					layer.bindPopup(featureName);
+					layer.on('mouseover', function() {
+						layer.openPopup();
+					});
+					layer.on('mouseout', function() {
+						layer.closePopup();
+					});
+					layer.on('click', function() {
```
mbucknell: This could be removed since the function is a noop.

src/main/webapp/js/views/MapView.js
```diff
@@ -39,10 +40,31 @@ define([
 				'World Physical': L.tileLayer.provider('Esri.WorldPhysical'),
 				'World Imagery' : L.tileLayer.provider('Esri.WorldImagery')
 			};
+			
+			// add public beaches
+			var yellowTriangleIcon = L.icon({
```
mbucknell: Consider moving your variable declarations to the beginning of the function. This is considered better style for javascript.

src/main/webapp/js/views/MapView.js
```diff
@@ -91,8 +113,9 @@ define([
 			this.map = L.map(MAP_DIV_ID, {
 				center: [41.0, -100.0],
 				zoom : 4,
-				layers : [this.baseLayers['World Street']]
+				layers : [this.baseLayers['World Street'], this.beachOverlays['Public Beaches'], this.beachOverlays['USGS Model Beaches']]
```
mbucknell: If you add the overlay layers here, aren't they visible when the map is first rendered? I just reread the JIRA and I think she just wants them added to the layer switcher. It's a little unclear but the acceptance criteria seems to say that they do not need to be on by default.

src/test/js/views/MapViewSpec.js
```diff
@@ -101,6 +101,13 @@ define([
 				testView.render();
 				expect(addControlSpy).toHaveBeenCalledWith(testView.defaultControls[0]);
 			});
+			
+			it('Expects layers to be created from geoJSON', function() {
+				testView.render();
+				expect(testView.addGeoJsonLayer).toHaveBeenCalledWith(jasmine.any(String), jasmine.any(L.Icon));
```
mbucknell: Maybe add a check that it gets called twice? Also addGeoJsonLayer actually gets called as part of initialization. It's actually important that they get called at initialization since the layers are added to the layer control before render is called.

Wasn't aware of the jasmine.any (etc) for matching parts of a result. These will be handy!

src/main/webapp/js/views/MapView.js
ayan-usgs: I think you're right. The acceptance criteria do imply that the layers are off by default rather that being on.

src/main/webapp/js/views/MapView.js
ayan-usgs: Good point.

ayan-usgs:
Implemented feedback from review.

PR #147, opened 2016-12-16
Enddat 270

mbucknell:


mbucknell:
Now the nwis site service is no longer proxied.

PR #155, opened 2017-01-18
Added EC to Data Sources and Partners

jlucido2:
I changed to 'Time Series' to be consistent with the labels in the rest of the app.


dblodgett-usgs:
:shipit:

PR #159, opened 2017-02-07
Added filter file formats to the user guide.

jlucido2:
I am pretty sure this is correct, but please review content and code of course.

### DIFF
--- a/src/main/webapp/UserGuide.jsp
+++ b/src/main/webapp/UserGuide.jsp
@@ -25,9 +25,8 @@
 				<h1>The Environmental Data Discovery and Transformation System</h1>
 
 				<p>
-					<em>This user guide is based on the re-implementation of the EnDDaT user interface. The latest functionality 
-						is available at: <a href="https://cida-test.er.usgs.gov/enddat/">https://cida-test.er.usgs.gov/enddat/</a> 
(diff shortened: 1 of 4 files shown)


src/main/webapp/UserGuide.jsp
```diff
@@ -25,9 +25,9 @@
 				<h1>The Environmental Data Discovery and Transformation System</h1>
 
 				<p>
-					<em>This user guide is based on the BETA re-implementation of the EnDDaT user interface. The latest functionality 
-						is available at: <a href="https://cida-test.er.usgs.gov/enddatv2/">https://cida-test.er.usgs.gov/enddatv2/</a> 
-						and production is available at: <a href="https://cida.usgs.gov/enddatv2/">https://cida.usgs.gov/enddatv2/</a></em>
+					<em>This user guide is based on the re-implementation of the EnDDaT user interface. The latest functionality 
+						is available at: <a href="https://cida-test.er.usgs.gov/enddat/">https://cida-test.er.usgs.gov/enddat/</a>
```
mbucknell: Is there a reason we mention the cida-test url? This does not seem like a good idea. The QA url should be used by cooperators only in my opinion.

jlucido2: Good point. I removed it. I don't really know why it was there in the first place.

PR #163, opened 2017-02-15
Add link to the filter file part of the documentation from the data processing part of the 'discovery' page

ayan-usgs:


src/main/webapp/js/hb_templates/processData.hbs
```diff
@@ -56,6 +56,7 @@
 				<!-- The file input field used as target for the file upload widget -->
 				<input id="time-filter-file-input" type="file" name="fileName">
 			</span>
+			<span>&nbsp;<a href="userguide#dataProcessing"">?</a></span>
```
mbucknell: You might consider using a font-awesome icon (maybe see how it looks), the is question, question-circle and question-circle-o

ayan-usgs: Thanks for the suggestion, I've give those icons a look.

